PR #418, opened 2023-05-03
Adjust string value extraction in ASCIIValue

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Adjust string value extraction in ASCIIValue to match PHP docs</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

According to [PHP docs about `gettype()`](https://www.php.net/manual/en/function.gettype.php#refsect1-function.gettype-returnvalues), it will never return a `false` value, but it will be `double` instead.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Formatter/ASCII/ASCIIValue.php
+++ b/src/core/etl/src/Flow/ETL/Formatter/ASCII/ASCIIValue.php
@@ -107,7 +107,6 @@ final class ASCIIValue
                 $this->stringValue = match (\gettype($val)) {
                     'string' => $val,
                     'boolean' => ($val) ? 'true' : 'false',
-                    /** @phpstan-ignore-next-line */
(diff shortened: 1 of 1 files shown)


stloyd:
In future, it may be useful to consider using [`get_debug_type()`](https://www.php.net/manual/en/function.get-debug-type.php) instead of `gettype()`.

PR #419, opened 2023-05-03
Modify `StrReplace` expression to work with arrays

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Modify `StrReplace` expression to work with arrays</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

PHP implementation of `str_replace` works both, with single strings, and arrays of pairs.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Reference/EntryExpression.php
+++ b/src/core/etl/src/Flow/ETL/Row/Reference/EntryExpression.php
@@ -257,7 +257,11 @@ trait EntryExpression
         return new Expressions(new StartsWith($this, $needle));
     }
 
-    public function strReplace(string $search, string $replace) : Expression
+    /**
+     * @param string|string[] $search
+     * @param string|string[] $replace
+     */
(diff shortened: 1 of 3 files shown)


norberttech:
I know its a bc break but maybe we should consider making arguments to also be an Expresssions? So you can str replace with values from other columns?

stloyd:
We can union 3 types IMO, but for expressions, we can add additional validation or force types.

norberttech:
on the DSL level we can use union types but in the expression itsefl I would probably go with Expressions only to reduce number of tests

PR #472, opened 2023-08-29
Allow to use Symfony UID for UUID generation

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Allow to use Symfony UID for UUID generation</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

The reason behind this is that now you can have either Ramsey or Symfony library, the `uuid()` will work with both.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Entry/UuidEntry.php
+++ b/src/core/etl/src/Flow/ETL/Row/Entry/UuidEntry.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Row\Entry;
+
+use Flow\ETL\Exception\InvalidArgumentException;
+use Flow\ETL\Row\Entry;
+use Flow\ETL\Row\Reference;
+use Flow\ETL\Row\Schema\Definition;
+
+/**
+ * @implements Entry<Entry\Type\Uuid, array{name: string, value: string}>
+ */
+final class UuidEntry implements \Stringable, Entry
+{
+    use EntryRef;
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public function __construct(private readonly string $name, private readonly Entry\Type\Uuid $value)
+    {
+        if ('' === $name) {
+            throw InvalidArgumentException::because('Entry name cannot be empty');
+        }
+    }
+
+    public static function from(string $name, string $value) : self
+    {
+        return new self($name, Entry\Type\Uuid::fromString($value));
+    }
+
+    public function __serialize() : array
+    {
+        return ['name' => $this->name, 'value' => $this->value->toString()];
+    }
+
+    public function __toString() : string
+    {
+        return $this->toString();
+    }
+
+    public function __unserialize(array $data) : void
+    {
+        $this->name = $data['name'];
+        $this->value = new Entry\Type\Uuid($data['value']);
+    }
+
+    public function definition() : Definition
+    {
+        return Definition::uuid($this->name);
+    }
+
+    public function is(string|Reference $name) : bool
+    {
+        if ($name instanceof Reference) {
+            return $this->name === $name->name();
+        }
+
+        return $this->name === $name;
+    }
+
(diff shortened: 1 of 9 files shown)


norberttech:
hey, one thing I would do is to add UUID entry type and let it hold some UUID interface that would allow for any of those implementations. 
You would need to adjust NativeEntryFactory and Cast expression to turn strings into UUIDs, and that should make it pretty flexible and easy to work with later

src/core/etl/src/Flow/ETL/Row/Entry/UuidEntry.php
```diff
@@ -0,0 +1,119 @@
...
+     * @throws InvalidArgumentException|RuntimeException
+     */
+    public function __construct(private readonly string $name, string|RamseyUuidInterface|SymfonyUuid $value)
+    {
+        if ('' === $name) {
+            throw InvalidArgumentException::because('Entry name cannot be empty');
+        }
+
+        if (\is_string($value)) {
+            if (\class_exists(RamseyUuid::class)) {
```
norberttech: I think this will throw an error when class does not exists, afaik you should use FQCN string instead here.

src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
```diff
@@ -289,6 +303,11 @@ private function isJson(string $string) : bool
         }
     }
 
+    private function isUuid(string $string) : bool
+    {
+        return 0 !== \preg_match('/\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\z/ms', $string);
```
norberttech: you came up with this regex or found it somewhere? Is it the same as symfony uid component is using?

src/core/etl/src/Flow/ETL/Row/Entry/UuidEntry.php
stloyd: Right, fixed.

src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
stloyd: It's taken from Ramsey library actually, I was not sure if we need to place a copy-right for it or not.

norberttech: Not sure from where Ramsey lib took it but I would extract it to const and put a reference that it was taken from Ramsey lib with credits

src/core/etl/src/Flow/ETL/Row/Entry/UuidEntry.php
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public function __construct(private readonly string $name, string|\Ramsey\Uuid\UuidInterface|\Symfony\Component\Uid\Uuid $value)
+    {
+        if ('' === $name) {
+            throw InvalidArgumentException::because('Entry name cannot be empty');
+        }
+
+        if (\is_string($value)) {
+            try {
+                if (\class_exists(\Ramsey\Uuid\UuidInterface::class)) {
```
norberttech: I think this might still fail (but you would need to test it), what I was using in other places was just a plain string `"\Ramsey\Uuid\UuidInterface"`

stloyd: Tested locally and with FCQN it works, tests are skipped properly.

PR #477, opened 2023-09-07
New PHAR build

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Added new Phar module that generates Flow-PHP phar file</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### Phar

1. Update dependencies: `composer install`,
2. Run the compilation step (be sure you have a `phar` extension enabled in your PHP binary): `composer build:compile`,
3. Test adjusted example with brand new phar: `./flow-php.phar examples/topics/aggregations/power_plant.php`

Expected result:
```shell
+---------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+-------------+
|    date | production_kwh_avg | consumption_kwh_avg | production_kwh_min | consumption_kwh_min | production_kwh_max | consumption_kwh_max | production_kwh_sum | consumption_kwh_sum | consumption |
+---------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+-------------+
| 2022/01 |              12.54 |               16.94 |               2.11 |               12.65 |              37.34 |               23.36 |             388.66 |               525.2 |     135.13% |
| 2022/02 |               24.5 |               17.75 |               5.41 |               14.76 |              48.16 |               29.04 |             685.93 |              497.11 |      72.47% |
| 2022/03 |              44.71 |               16.48 |               4.73 |               12.15 |              63.89 |               21.69 |            1386.03 |              510.81 |      36.85% |
| 2022/04 |              32.03 |               14.72 |               1.26 |                 3.5 |              57.02 |               22.11 |             448.42 |              206.09 |      45.96% |
+---------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+--------------------+---------------------+-------------+
4 rows
```

----

```shell
stloyd@MacBook-Pro-2 flow % ./flow-php.phar --version
Flow-PHP - Extract Transform Load - Data processing framework 1.0.x-02fba709
```


.github/workflows/build-executable.yml
```diff
@@ -0,0 +1,90 @@
...
+          workflow: "release-build.yml"
+          name: "spc-linux-x86_64"
+
+      - name: Validate SPC bin
+        run: |
+          chmod +x spc
+          ./spc --version
+
+      - name: "Download sources"
+        run: ./spc download --with-php=${{ matrix.php-version }} php-src,micro,pkg-config,onig,zlib,openssl,gmp,libiconv,libxml2
```
stloyd: We can add an additional cache layer for those files.

.github/workflows/build-executable.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: "Build phpmicro Executable"
+        run: |
+          SPC_USE_SUDO=yes ./spc doctor --auto-fix
+          ./spc build pcntl,posix,mbstring,tokenizer,dom,xmlreader,filter,openssl,bcmath,gmp,fileinfo,phar --build-micro
```
stloyd: When this PR (https://github.com/crazywhalecc/static-php-cli/pull/188) lands, we can greatly reduce the time of the build.

.github/workflows/build-executable.yml
stloyd: Added in: https://github.com/flow-php/flow/pull/477/commits/ab8130bccf63e36c9e969ec09c9306032b1edff2

build/runtime.php
```diff
@@ -0,0 +1,46 @@
...
+
+            return Command::FAILURE;
+        }
+
+        $pipeline = include $filename;
+        if (!$pipeline instanceof DataFrame) {
+            $style->error('Expecting Flow-PHP DataFrame, received: ' . (is_object($pipeline) ? get_class($pipeline) : gettype($pipeline)));
+
+            return Command::FAILURE;
+        }
```
norberttech: This logic could be extracted to standalone DataFrameLoader which can be properly covered with tests

build/runtime.php
```diff
@@ -0,0 +1,46 @@
...
+        $style = new SymfonyStyle($input, $output);
+
+        $filename = (string) $input->getArgument('input-file');
+        if (!\file_exists($filename)) {
+            $style->error("Input file ($filename) doesn't exists!");
+
+            return Command::FAILURE;
+        }
+
+        if (!\str_ends_with($filename, '.php')) {
```
norberttech: I think I would read the first 5 bytes of the file and check if it starts from `<?php` string (lower/upper case)

build/runtime.php
norberttech: once this is extracted to standalone class (it can be passed by constructor with default implementation) you can also put a simple test for this command and pass DataFrameLoader as a mock

build/version.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+final class FlowVersion
+{
+    public static function getVersion(): string
+    {
+        return '1.0.x-@git_commit_short@';
```
norberttech, line 9: Could you elaborate on this one? I'm not sure if I understand why it's needed

.github/workflows/build-executable.yml
norberttech: any chance to put snappy extension here?

norberttech:
is there a way to reproduce the same process from github actions workflow locally without using act to emulate actions?

build/runtime.php
stloyd: Where do you see this new class? Cause it's not a real loader like we have already in ETL, so implementing the `Loader` interface seems wrong (as we don't have context nor rows).

build/version.php
stloyd: It allows us to set the version of build phar (and binary in the end), so we could build them based on the `1.x` branch as well on other versioning, it was just an idea to match the current branch name and add git commit into the name of the version, I'm open to change this.

build/runtime.php
norberttech: ah sorry I wasn't thinking about making it an ETL Loader, it's just a name, could be also DataFrameFactory or anything similar

build/version.php
norberttech: but is the `git_commit_short` a placeholder or just a hardcoded value? The only thing I'm worried if this value is hardcoded is that we will never remember to update it while releasing new versions.

.github/workflows/build-executable.yml
stloyd: Added but seems like there is a bug to build snappy on Ubuntu, I will report later to the library.

stloyd:
Added notes to the description, should be enough to generate the binary locally.

build/version.php
stloyd: It's a placeholder for the Box library, when compiling phar it's replaced with a proper git commit.

Refs: https://box-project.github.io/box/configuration/#replaceable-placeholders

norberttech:
> Added notes to the description, should be enough to generate the binary locally.

I can't see it, did you maybe forgot to push the commit?

build/runtime.php
stloyd: Done.

.github/workflows/build-executable.yml
stloyd: Error for Snappy + Ubuntu: https://github.com/crazywhalecc/static-php-cli/issues/195

.github/workflows/build-executable.yml
```diff
@@ -0,0 +1,129 @@
+name: Build Flow-PHP executable
+
+on:
+  pull_request:
```
norberttech: do we need to build execs at every PR? Maybe we can schedule it for daily and then for releases?

stloyd: I would leave 2-3 options (this one is for testing), merge to the main branch, release and maybe manual trigger, or just the last two.

PR #554, opened 2023-10-09
Adjust `size` expression to behave similarly to the `count` one

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Adjust the `size` expression to behave similarly to the `count` one</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Now, the `size` expression will return `null` when calling it on non-string & non-countable values, similarly to what's done for the `count` expression.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Reference/Expression/Size.php
+++ b/src/core/etl/src/Flow/ETL/Row/Reference/Expression/Size.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Flow\ETL\Row\Reference\Expression;
 
-use Flow\ETL\Exception\RuntimeException;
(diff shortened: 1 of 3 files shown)


norberttech:
That should be default behavior for any expression, if execution is not possible on given cell, it should return null 👏

PR #591, opened 2023-10-16
Skip JSON & XML checks for string that doesn't look as this type of data

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Skip JSON & XML checks for a string that doesn't look like this type of data</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Thanks to Blackfire, I can confirm that adding simple string checks for JSON & XML improves the performance of `NativeEntityFactory`.

Zrzut ekranu 2023-10-16 o 14 20 20

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
+++ b/src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
@@ -54,15 +54,17 @@ final class NativeEntryFactory implements EntryFactory
 
         if (\is_string($value)) {
             if ('' !== $value) {
-                if ($this->isJson($value)) {
+                $trimmedValue = \trim($value);
+
+                if ($this->isJson($trimmedValue)) {
                     return Row\Entry\JsonEntry::fromJsonString($entryName, $value);
                 }
 
-                if ($this->isUuid($value)) {
+                if ($this->isUuid($trimmedValue)) {
                     return new Row\Entry\UuidEntry($entryName, Entry\Type\Uuid::fromString($value));
                 }
 
-                if ($this->isXML($value)) {
+                if ($this->isXML($trimmedValue)) {
                     return new Entry\XMLEntry($entryName, $value);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
```diff
@@ -330,6 +334,10 @@ private function isUuid(string $string) : bool
 
     private function isXML(string $string) : bool
     {
+        if ('<' !== $string[0]) {
```
norberttech, line 339: I think we might want trim string just in case of some empty spaces at the beginning.

src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
```diff
@@ -304,6 +304,10 @@ private function getClass(object $object) : string
 
     private function isJson(string $string) : bool
     {
+        if ('{' !== $string[0] && '[' !== $string[0]) {
```
norberttech, line 309: just like with the xml, I think we should trim string first.

src/core/etl/src/Flow/ETL/Row/Factory/NativeEntryFactory.php
stloyd: But do you mean in all of that check or globally? Like `$value = \trim($value)` before passing it to any of the checks:
```php
        if (\is_string($value)) {
            $value = \trim($value);

            if ('' !== $value) {
                if ($this->isJson($value)) {
                // ...

norberttech: `$this->isXML(trim($string))` - we don't want to trim values that should not be trimmed as spaces at the beginning might be expected in some scenarios

PR #598, opened 2023-10-16
Rework GH testing actions to run only after code changes

### LINKED ISSUE #538: Run tests only after code changes
Currently, whole test suite is executed at every pull request however, we are doing some PR's that are not affecting the codebase, for example: 

- updating docs
- adding examples

It would be nice to execute tests only when code/dependencies were changed.

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Rework GH testing actions to run only after code changes</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Fixes #538

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -2,9 +2,15 @@ name: Test Suite
 
 on:
   pull_request:
+    paths:
+      - '.github/workflows/**'
+      - 'src/adapter/**'
+      - 'src/core/**'
+      - 'src/lib/**'
+      - 'tools/**'
+      - 'composer.lock'
   push:
     branches: [ 1.x ]
-  schedule: [ { cron: '0 8 * * *' } ]
 
 jobs:
   static-analyze:
@@ -29,7 +35,7 @@ jobs:
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          coverage: pcov
+          coverage: none
           tools: composer:v2
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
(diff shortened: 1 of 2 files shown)


.github/workflows/test-suite.yml
```diff
@@ -2,9 +2,15 @@ name: Test Suite
 
 on:
   pull_request:
+    paths:
+      - '.github/workflows/**'
```
norberttech, line 6: do we need it at all workflows changes, or just `test-suite.yml` here?

.github/workflows/test-benchmark.yml
```diff
@@ -2,6 +2,12 @@ name: Benchmark Suite
 
 on:
   pull_request:
+    paths:
```
norberttech, line 5: no workflows here?

.github/workflows/test-suite.yml
```diff
@@ -29,7 +35,7 @@ jobs:
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          coverage: pcov
+          coverage: none
```
norberttech, line 38: nice, that should speedup test suite a bit 👍

.github/workflows/test-suite.yml
stloyd: This was done because a test workflow is required for pull requests so it was hanging with a status pending

.github/workflows/test-benchmark.yml
stloyd: We can add it but as mentioned there is a workaround not as a requirement

PR #599, opened 2023-10-16
Reduced total available memory to make sure that MemorySort is doing a fallback to cache

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Reduced total available memory to make sure that MemorySort is doing a fallback to cache</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
It seems that this can happen when current memory consumption is not exceeding the available limit, then whole sorting occurs in memory, and cache is not used at all, and the failing test is about the cache usage. 
What I still don't understand is why it started just now, but it might be related to optimizations of NativeEntryFactory.

PR #611, opened 2023-10-18
Added RLEBigPackingHybrid encode function

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>RLEBigPackingHybrid encode function</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

We need this to implement parquet file writer, it's also a pretty cool standalone algo that can be used to reduce the amount of space needed for storing integers. 

More details here: https://parquet.apache.org/docs/file-format/data-pages/encodings/#RLE

I also slightly improved reading RLE_DICTIONARY encoded data pages by simplifying reading conditions that relays on the expected number of items rather than data length and item numbers.

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
+++ b/src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
@@ -1,0 +1,119 @@
+<?php declare(strict_types=1);
+
+namespace Flow\Parquet\BinaryWriter;
+
+use Flow\Parquet\BinaryWriter;
+use Flow\Parquet\ByteOrder;
+use Flow\Parquet\DataSize;
+
+final class BinaryBufferWriter implements BinaryWriter
+{
+    private DataSize $length;
+
+    public function __construct(private string &$buffer, private readonly ByteOrder $byteOrder = ByteOrder::LITTLE_ENDIAN)
+    {
+        $this->buffer = '';
+        $this->length = new DataSize(0);
+    }
+
+    public function length() : DataSize
+    {
+        return $this->length;
+    }
+
+    public function writeBits(array $bits) : void
+    {
+        $byte = 0;
+        $bitIndex = 0;
+
+        foreach ($bits as $bit) {
+            if ($bit) {
+                $byte |= (1 << $bitIndex);
+            }
+
+            $bitIndex++;
+
+            if ($bitIndex === 8) {
+                $this->buffer .= \chr($byte);
+                $this->length->addBytes(1); // Assume addBytes is a method to add to the length
+                $byte = 0;
+                $bitIndex = 0;
+            }
+        }
+
+        // If there are remaining bits that don't fill a byte
+        if ($bitIndex > 0) {
+            $this->buffer .= \chr($byte);
+            $this->length->addBytes(1);
+        }
+    }
+
+    public function writeBooleans(array $values) : void
+    {
+        $bits = [];
+
+        foreach ($values as $value) {
+            $bits[] = $value ? 1 : 0;
+        }
+        $this->writeBits($bits);
+    }
+
+    public function writeBytes(array $bytes) : void
+    {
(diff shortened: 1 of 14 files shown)


src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
```diff
@@ -0,0 +1,119 @@
...
+    public function writeStrings(array $strings) : void
+    {
+        $format = $this->byteOrder === ByteOrder::BIG_ENDIAN ? 'N' : 'V';
+
+        foreach ($strings as $string) {
+            $length = \strlen($string);
+            $this->buffer .= \pack($format, $length);
+            $this->buffer .= $string;
+        }
+        $this->length->addBytes(\array_sum(\array_map('strlen', $strings)) + (4 * \count($strings)));
```
stloyd, line 101: If we store the length and count it in the loop above we could remove array_map

src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public function writeInts64(array $ints) : void
+    {
+        $format = $this->byteOrder === ByteOrder::BIG_ENDIAN ? 'J' : 'P';
+
+        foreach ($ints as $int) {
+            $this->buffer .= \pack($format, $int);
+        }
+        $this->length->addBytes(\count($ints) * 8);
```
stloyd, line 86: Count could be replaced with math in the above loop

src/lib/parquet/src/Flow/Parquet/ParquetFile/Data/RLEBitPackedHybrid.php
```diff
@@ -112,7 +111,115 @@ public function decodeRLE(BinaryReader $reader, int $bitWidth, int $intVar, int
...
+            $previousValue = $value;
+        }
+
+        if (\count($rleBuffer) > 8) {
+            $this->encodeRLE($writer, $bitWidth, $rleBuffer);
+            $rleBuffer = [];
+        }
+
+        if (\count($bitPackedBuffer)) {
+            if (\count($rleBuffer)) {
```
stloyd, line 196: Probably can be removed

src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
norberttech: that would be slower since it would do the same calculation `\count($ints)` times instead of doing it once.

src/lib/parquet/src/Flow/Parquet/ParquetFile/Data/RLEBitPackedHybrid.php
norberttech: it can't, if there is a situation where rleBuffer is not full, there are no more values and bitPackedBuffer is also not empty, in that case, we would lose those last values.

src/lib/parquet/src/Flow/Parquet/BinaryWriter/BinaryBufferWriter.php
norberttech: I'm gonna get back to optimizations once I finish the writer, while working on the reader, I spent a good amount of time optimizing things that a few days later were removed/replaced with something else, I don't want to make that mistake again.

PR #619, opened 2023-10-20
Move implementation of entry structure creation into the `NativeEntryFactory`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Move implementation of entry structure creation into the `NativeEntryFactory`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
I would actually like to keep them as a simple functions if thats not creating any additional complications. This way they are more reusable and I already have some use cases in mind

stloyd:
Structure building right now is the only one that is done completely differently than others.

This also is a little confusing cause that function requires still having an entry factory.

norberttech:
> This also is a little confusing cause that function requires still having an entry factory.

Valid point, it creates a weird dependency 👍

PR #620, opened 2023-10-20
Reduce CI load by cancelling previous runs with every new commit in PR

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Reduce CI load by cancelling previous runs with every new commit in PR</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Based on: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/72408109#72408109

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,9 +2,19 @@ name: Documentation Linter
 
 on:
   pull_request:
+    paths:
+      - 'src/**'
+      - 'docs/**'
+      - 'examples/**'
+      - 'README.md'
   push:
     branches:
       - "1.x"
+
+# See https://stackoverflow.com/a/72408109
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   documentation-linter:
--- a/.github/workflows/test-benchmark.yml
+++ b/.github/workflows/test-benchmark.yml
@@ -11,6 +11,11 @@ on:
       - 'composer.json'
       - 'composer.lock'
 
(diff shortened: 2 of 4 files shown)


.github/workflows/docs.yml
```diff
@@ -2,10 +2,20 @@ name: Documentation Linter
 
 on:
   pull_request:
+    paths:
```
stloyd, line 5: This change requires removing that workflow from required as it's not being run right now.

norberttech:
hmm I'm getting job cancellation notifications, isn't that conflicting with post merge jobs also?

stloyd:
Maybe just turn off those cause they are not useful as cancellation should happen on manual trigger or particular failure of infrastructure.

norberttech:
oh I know what's happening, whenver we merge anything, automation is also pushing changes to the CHANGELOG.md file, which triggers all tests all over again. Could you take a look into this and try to figure out if there is any way to prevent triggering the whole testsuite after the update of CHANGELOG.md?

stloyd:
@norberttech Should be fixed with: https://github.com/flow-php/flow/pull/623

PR #625, opened 2023-10-22
Remove `array_merge()` from constructors when not needed

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Remove `array_merge()` from constructors when not needed</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Join/Comparison/All.php
+++ b/src/core/etl/src/Flow/ETL/Join/Comparison/All.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Flow\ETL\Join\Comparison;
 
+use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\Join\Comparison;
 use Flow\ETL\Row;
 use Flow\ETL\Row\EntryReference;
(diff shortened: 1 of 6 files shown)


src/core/etl/src/Flow/ETL/Join/Comparison/All.php
```diff
@@ -18,9 +19,13 @@ final class All implements Comparison
      */
     private array $comparisons;
 
-    public function __construct(Comparison $comparison, Comparison ...$comparisons)
+    public function __construct(Comparison ...$comparisons)
```
norberttech, line 22: I prefer to keep it like it is now. This way, the constructor enforces at least one Comparison, so there is no need to write a unit test covering this edge case to satisfy mutation tests. 
It's an approach commonly used in Scala when at least one element is required. 
This particular array_merge should also have no impact on the pipeline processing performance.

PR #628, opened 2023-10-22
Remove dead `array_merge_recursive()` function from Parquet

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Remove dead `array_merge_recursive()` function from Parquet</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

There is a native function in PHP: https://www.php.net/manual/en/function.array-merge-recursive.php, as well this code is never called in the codebase.

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/functions.php
+++ b/src/lib/parquet/src/Flow/Parquet/functions.php
@@ -5,26 +5,6 @@ declare(strict_types=1);
 namespace Flow\Parquet;
 
 /**
- * @psalm-suppress MixedAssignment
- * @psalm-suppress DocblockTypeContradiction
- * @psalm-suppress MixedArrayOffset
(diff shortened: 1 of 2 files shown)


norberttech:
the native function does not work the same way as this one, I would like to keep it for a little longer since I still might need it while working on the writer.

PR #650, opened 2023-10-26
Run cs-fixer against codebase

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Run cs-fixer against codebase</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

This is a step before the cleanup of rules & sets definitions in the cs-fixer configuration.

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/Reader.php
+++ b/src/lib/parquet/src/Flow/Parquet/Reader.php
@@ -22,7 +22,7 @@ final class Reader
             throw new InvalidArgumentException("File {$path} does not exist");
         }
 
-        $stream = \fopen($path, 'r');
+        $stream = \fopen($path, 'rb');
 
         if (!\is_resource($stream)) {
             throw new InvalidArgumentException("File {$path} is not a valid resource");
--- a/src/lib/parquet/src/Flow/Parquet/Reader.php
+++ b/src/lib/parquet/src/Flow/Parquet/Reader.php
@@ -22,7 +22,7 @@ final class Reader
             throw new InvalidArgumentException("File {$path} does not exist");
         }
 
-        $stream = \fopen($path, 'rb');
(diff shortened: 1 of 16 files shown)


src/lib/parquet/src/Flow/Parquet/Writer.php
```diff
@@ -30,7 +30,7 @@ public function write(string $path, Schema $schema, iterable $rows) : void
             \unlink($path);
         }
 
-        $stream = \fopen($path, 'wb');
+        $stream = \fopen($path, 'w');
```
norberttech: I think this should stay wb

src/lib/parquet/src/Flow/Parquet/Reader.php
```diff
@@ -22,7 +22,7 @@ public function read(string $path) : ParquetFile
             throw new InvalidArgumentException("File {$path} does not exist");
         }
 
-        $stream = \fopen($path, 'rb');
+        $stream = \fopen($path, 'r');
```
norberttech: same here

src/lib/parquet/src/Flow/Parquet/Writer.php
stloyd: Restored, but it's now added elsewhere as it's not configurable other way... https://cs.symfony.com/doc/rules/function_notation/fopen_flags.html

PR #656, opened 2023-10-27
Adjust `Rows::chunk()` to work on generators instead of arrays

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Adjust Rows::chunk() to work on generators instead of arrays</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Old:
```console
+-----------------------------+--------------------------+------+-----+-----------+-----------+
| benchmark                   | subject                  | revs | its | mem_peak  | mode      |
+-----------------------------+--------------------------+------+-----+-----------+-----------+
| RowsBench                   | bench_chunk_10           | 5    | 5   | 60.408mb  | 3.335ms   |
+-----------------------------+--------------------------+------+-----+-----------+-----------+
```

New:
```console
+-----------------------------+--------------------------+------+-----+-----------+-----------+
| benchmark                   | subject                  | revs | its | mem_peak  | mode      |
+-----------------------------+--------------------------+------+-----+-----------+-----------+
| RowsBench                   | bench_chunk_10           | 5    | 5   | 57.885mb  | 1.600μs   |
+-----------------------------+--------------------------+------+-----+-----------+-----------+
```

### DIFF
--- a/src/core/etl/src/Flow/ETL/ExternalSort/CacheExternalSort.php
+++ b/src/core/etl/src/Flow/ETL/ExternalSort/CacheExternalSort.php
@@ -34,7 +34,7 @@ final class CacheExternalSort implements ExternalSort
     {
         /** @var array<string, \Generator<Rows>> $cachedPartsArray */
         $cachedPartsArray = [];
-        $maxRowsSize = 0;
+        $maxRowsSize = 1;
 
         /** @var int $i */
         foreach ($this->cache->read($this->id) as $i => $rows) {
--- a/src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
+++ b/src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
@@ -31,7 +31,7 @@ final class MemorySort implements ExternalSort
         private readonly Cache $cache,
         private Unit $maximumMemory
     ) {
-        $this->configuration = new Configuration($safetyBufferPercentage = 10);
+        $this->configuration = new Configuration(10);
 
         if ($this->configuration->isLessThan($maximumMemory) && !$this->configuration->isInfinite()) {
             /**
@@ -48,7 +48,7 @@ final class MemorySort implements ExternalSort
         $memoryConsumption = new Consumption();
 
         $mergedRows = new Rows();
-        $maxSize = 0;
+        $maxSize = 1;
 
         foreach ($this->cache->read($this->cacheId) as $rows) {
             $maxSize = \max($rows->count(), $maxSize);
@@ -64,6 +64,6 @@ final class MemorySort implements ExternalSort
 
         $this->cache->clear($this->cacheId);
 
-        return new Extractor\ProcessExtractor(...$mergedRows->sortBy(...$refs)->chunks($maxSize));
+        return new Extractor\ProcessExtractor(...\iterator_to_array($mergedRows->sortBy(...$refs)->chunks($maxSize)));
     }
 }
--- a/src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
+++ b/src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
@@ -64,6 +64,6 @@ final class MemorySort implements ExternalSort
 
         $this->cache->clear($this->cacheId);
 
-        return new Extractor\ProcessExtractor(...\iterator_to_array($mergedRows->sortBy(...$refs)->chunks($maxSize)));
+        return new Extractor\GeneratorExtractor($mergedRows->sortBy(...$refs)->chunks($maxSize));
     }
 }
--- a/src/core/etl/src/Flow/ETL/Extractor/GeneratorExtractor.php
+++ b/src/core/etl/src/Flow/ETL/Extractor/GeneratorExtractor.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Flow\ETL\Extractor;
 
+use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\Extractor;
 use Flow\ETL\FlowContext;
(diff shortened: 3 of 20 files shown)


src/core/etl/src/Flow/ETL/ExternalSort/CacheExternalSort.php
```diff
@@ -34,7 +34,7 @@ public function sortBy(EntryReference ...$refs) : Extractor
     {
         /** @var array<string, \Generator<Rows>> $cachedPartsArray */
         $cachedPartsArray = [];
-        $maxRowsSize = 0;
+        $maxRowsSize = 1;
```
norberttech, line 37: could you elaborate on this one? I don't remember anymore that logic and would need to go through the code again to understand why it was set to 0 and why it's even needed

stloyd: This code was IMO false positive, cause as the chunk size was always: `int<1, max>`, and never `0`, this value was always overwritten by a later loop which has: `$maxRowsSize = \max($maxRowsSize, $rows->count());`.

Now with a clear declaration of what chunk size is, it was reported by static analysis that this value is wrong & should never happen.

src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
```diff
@@ -64,6 +64,6 @@ public function sortBy(EntryReference ...$refs) : Extractor
 
         $this->cache->clear($this->cacheId);
 
-        return new Extractor\ProcessExtractor(...$mergedRows->sortBy(...$refs)->chunks($maxSize));
+        return new Extractor\ProcessExtractor(...\iterator_to_array($mergedRows->sortBy(...$refs)->chunks($maxSize)));
```
norberttech: hmm, can't we return here `$mergedRows->sortBy(...$refs)->chunks($maxSize)` which is now a generator?

stloyd: You cannot unpack a generator in PHP 😞

src/core/etl/tests/Flow/ETL/Tests/Benchmark/RowsBench.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types=1);
+
+use Flow\ETL\Rows;
+use PhpBench\Attributes\Groups;
+use PhpBench\Attributes\Iterations;
+use PhpBench\Attributes\Revs;
+
+#[Iterations(5)]
+#[Groups(['rows'])]
```
stloyd: @norberttech Not sure if we should add a new group or re-use some existing ones.

src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
norberttech: But it should be pretty straightforward to create GeneratorExtractor that would accept a Generator in the constructor and just use it to yield rows in the extract method which I think should reduce the memory consumption, no?

src/core/etl/tests/Flow/ETL/Tests/Benchmark/RowsBench.php
norberttech: I think I would rename entry_factory into `building_blocks` and put NativeEntryfactory and Rows benchmarks under it, wdyt?

src/core/etl/src/Flow/ETL/ExternalSort/MemorySort.php
stloyd: Added new extractor.

src/core/etl/src/Flow/ETL/Extractor/GeneratorExtractor.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * @param \Generator<Rows> $rows
+     */
+    public function __construct(private readonly \Generator $rows)
+    {
+    }
+
+    public function extract(FlowContext $context) : \Generator
+    {
+        yield from $this->rows;
```
norberttech: I wonder if we should not actually iterate and just make sure that each result from inner Generator is an instance of Rows

stloyd: I'm not sure if it would not be overkill, cause this class is marked as an internal one (same as `ProcessExtractor`), and it's used only in `MemorySort` which provides a `Rows` class.

src/core/etl/tests/Flow/ETL/Tests/Benchmark/RowsBench.php
stloyd: Sounds good I guess, more general than just having `entry_factory` as one of the smallest groups.

src/core/etl/src/Flow/ETL/Extractor/GeneratorExtractor.php
norberttech: But since PHP does not have private classes, nothing prevents anyone from using it, and additional iteration should not affect performance more than the previous itereator_to_array solution

stloyd: I still think it's overkill, but added such a check.

PR #657, opened 2023-10-28
Fixed parquet writer performance degradation

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>parquet writer performance degradation</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/ThriftStream/TPhpFileStream.php
+++ b/src/lib/parquet/src/Flow/Parquet/ThriftStream/TPhpFileStream.php
@@ -3,7 +3,6 @@
 namespace Flow\Parquet\ThriftStream;
 
 use Thrift\Exception\TException;
-use Thrift\Factory\TStringFuncFactory;
 use Thrift\Transport\TTransport;
 
 final class TPhpFileStream extends TTransport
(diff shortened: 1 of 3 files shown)


src/lib/parquet/src/Flow/Parquet/ThriftStream/TPhpFileStream.php
```diff
@@ -63,15 +62,16 @@ public function read($len)
 
     public function write($buf) : void
     {
-        while (TStringFuncFactory::create()->strlen($buf) > 0) {
+        while ($buf != '') {
```
norberttech, line 65: I just noticed, how terrible those default classes from Thrift 🤦 instead of using strlen, they are creating an object that is using strlen inside 🤦 🤦 🤦  I might need to look into other classes and create our own custom implementations just slightly optimized or maybe even optimize it directly in thrift repo

stloyd: Shouldn't this be strict check?

norberttech: ah shit 🤦 IDE replaced this for me and I didn't notice it's != instead of !==. It wont affect anything as there can't be anything different than string here, will correct that later. Do you know if there is a rule in phpstan/psalm maybe that could enforce strict checks?

stloyd: There is a cs-fixer rule: https://cs.symfony.com/doc/rules/strict/strict_comparison.html but is really dumb one, and will lead to replacing every `==` with `===` without checking the code, so objects comparison will be also replaced will fail after applying.

PR #659, opened 2023-10-28
Prevent calling same class methods inside of schema files

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Prevent calling same class methods inside of schema files</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Skipping calling additional methods in the matter of accessing internal variables in class improves performance.

See: https://3v4l.org/sOIct

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/ParquetFile/Schema/FlatColumn.php
+++ b/src/lib/parquet/src/Flow/Parquet/ParquetFile/Schema/FlatColumn.php
@@ -150,8 +150,8 @@ final class FlatColumn implements Column
     {
         return [
             /** @phpstan-ignore-next-line */
-            'type' => $this->type()->name . ($this->logicalType()?->name() !== null ? ' (' . $this->logicalType()?->name() . ')' : ''),
-            'optional' => $this->repetition()?->value === Repetition::OPTIONAL->value,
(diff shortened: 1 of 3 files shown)


norberttech:
Hmm not sure if I get the intention of this, is that affecting performance in any way?

stloyd:
@norberttech It was late when I pushed this 😂 Added comment into description:

> Skipping calling additional methods in the matter of accessing internal variables in class improves performance.
> See: [3v4l.org/sOIct](https://3v4l.org/sOIct)

norberttech:
I would rather postpone this adjustment for now, there are two reasons: 

- it does not seem to be impactful enough to bring a visible value to the performance at this state 
- For some of those methods I'm planning to make mnemonics that will execute only once (like flatPath), and then return the result like, for example isList or isListElement, that's why I'm using methods directly rather than properties

PR #676, opened 2023-10-30
Improve performance for `Rows`: `dropRight`, `partitionBy` & `sortBy`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Improve performance for `Rows`: `dropRight`, `partitionBy` & `sortBy`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Rows.php
+++ b/src/core/etl/src/Flow/ETL/Rows.php
@@ -134,13 +134,7 @@ final class Rows implements \ArrayAccess, \Countable, \IteratorAggregate, Serial
 
     public function dropRight(int $size) : self
     {
-        $rows = $this->rows;
-
-        for ($i = 0; $i < $size; $i++) {
-            \array_pop($rows);
-        }
-
-        return new self(...$rows);
+        return new self(...\array_slice($this->rows, 0, -$size));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/core/etl/src/Flow/ETL/Rows.php
```diff
@@ -226,13 +212,7 @@ public function first() : Row
...
     public function flatMap(callable $callable) : self
     {
-        $rows = [];
-
-        foreach ($this->rows as $row) {
-            $rows[] = $callable($row);
-        }
-
-        return new self(...\array_merge(...$rows));
+        return new self(...\array_merge(...\array_map($callable, $this->rows)));
```
norberttech: afaik on bigger datasets, foreach is faster than array_map, I believe that if you look into commits history you will notice that it was initially done like that, but then I reverted it into foreach - array_map looks to be more memory efficient but slower than foreach

stloyd: Reverted as with more revs count in benchmark it was slower, while giving ~2% memory boost.

norberttech (approved):

PR #693, opened 2023-10-31
Added library version to parquet created_by metadata

### LINKED ISSUE #575: Parquet writer
This is a continuation of #506 
The goal is to build a parquet writer that: 

- supports all parquet types
- allows appending data into an existing file
- supports remote filesystems 

This is an absolute minimum to replace the library flow is using right now. After that the next step is to work on more direct integration between Flow and Parquet in order to benefit from more advanced parquet features not available in other file formats.

### LINKED ISSUE #691: Parquet - add library name/version to writer
Writer should put the following string into metadata created_by field: 

```
flow-php parquet version 0.5.0
```

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Added library version to parquet created_by metadata</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes: #691
Closes: #575

### DIFF
--- a/src/lib/parquet/composer.json
+++ b/src/lib/parquet/composer.json
@@ -15,6 +15,7 @@
         "ext-bcmath": "*",
         "ext-zlib": "*",
         "psr/log": "^2.0 || ^3.0",
+        "composer-runtime-api": "^2.1",
         "flow-php/dremel": "1.x-dev",
         "flow-php/snappy": "1.x-dev"
     },
--- a/src/lib/parquet/composer.json
+++ b/src/lib/parquet/composer.json
@@ -14,6 +14,7 @@
         "php": "~8.1 || ~8.2",
         "ext-bcmath": "*",
         "ext-zlib": "*",
+        "psr/log": "^2.0 || ^3.0",
         "flow-php/dremel": "1.x-dev",
         "flow-php/snappy": "1.x-dev"
     },
--- a/src/lib/parquet/src/Flow/Parquet/Writer.php
+++ b/src/lib/parquet/src/Flow/Parquet/Writer.php
@@ -2,6 +2,7 @@
 
 namespace Flow\Parquet;
 
+use Composer\InstalledVersions;
 use Flow\Parquet\Data\DataConverter;
 use Flow\Parquet\Exception\InvalidArgumentException;
 use Flow\Parquet\Exception\RuntimeException;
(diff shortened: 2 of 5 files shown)


src/lib/parquet/src/Flow/Parquet/Writer.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Flow\Parquet;
 
+use Composer\InstalledVersions;
```
stloyd, line 5: Adding this class means that `composer.json` needs to have a clear dependency on `composer-runtime-api:^2.0` both here and in the main folder.

stloyd (changes requested):

src/lib/parquet/src/Flow/Parquet/Writer.php
norberttech: Could you elaborate? I thought this comes with the composer itself and it's always available through autoload. What exactly is the link to that package you mentioned on packagist?

stloyd: It's true, but from the perspective of dependencies, this class doesn't exist till you use Composer. It's not a real package, it's a virtual one but allows marking it as a clear dependency, similar fix here: https://github.com/elastic/elastic-transport-php/issues/8#issuecomment-1211937261

src/lib/parquet/composer.json
```diff
@@ -14,6 +14,8 @@
         "php": "~8.1 || ~8.2",
         "ext-bcmath": "*",
         "ext-zlib": "*",
+        "psr/log": "^2.0 || ^3.0",
+        "composer-runtime-api": "^2.1",
         "flow-php/dremel": "1.x-dev",
         "flow-php/snappy": "1.x-dev"
```
stloyd, line 20:
```suggestion
        "composer-runtime-api": "^2.1",
        "flow-php/dremel": "1.x-dev",
        "flow-php/snappy": "1.x-dev",
        "psr/log": "^2.0 || ^3.0"
```

stloyd (approved):

PR #694, opened 2023-10-31
Replaced codename parquet with flow parquet library in parquet adapter

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Replaced codename parquet with flow parquet library in parquet adapter</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Codename Parquet library dependency</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-parquet/src/Flow/ETL/Adapter/Parquet/ParquetLoader.php
```diff
@@ -14,31 +10,28 @@
...
 
     private ?Schema $inferredSchema = null;
 
-    private ?ParquetDataWriter $writer = null;
+    private ?Writer $writer = null;
 
     public function __construct(
         private readonly Path $path,
-        private readonly int $rowsInGroup = 10000,
+        private readonly Options $options,
```
stloyd, line 31: This change requires to be applied also on examples & benchmark files.

norberttech: It was applied on the benchmarks and examples, the reason why benchmarks are not working is related to the fact that NativeEntryFactory while reading the parquet file is identifying uuid strings and creating UuidEntry which returns Uuid object when Rows::toArray is called since it's just iterating over entries and calling `$entry->value()` on it.

PR #699, opened 2023-10-31
Add a new `MapEntry` type

### LINKED ISSUE #684: MapEntry - new entry type
Map is a Key/Value data structure where each key must be unique. 
Keys can have any of the other Flow simple data types like: 

- String
- Int
- DateTime
- Float
- Uuid

Map Value can be literally anything, it's just an Entry. 

<s>Something like: `Entry::map(string $name, Entry $key, Entry $value)` </s>

Something like: `Entry::map(string $name, array $map)`

This would let us create even stronger integration with Parquet and Avro and other data destinations that support Maps.  

One important note on a map is that all keys must have exactly the same type, so we can't have mixed int|string keys.

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `MapEntry` type</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes #684

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Entry/MapEntry.php
+++ b/src/core/etl/src/Flow/ETL/Row/Entry/MapEntry.php
@@ -1,0 +1,151 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Row\Entry;
+
+use Flow\ArrayComparison\ArrayComparison;
+use Flow\ETL\Exception\InvalidArgumentException;
+use Flow\ETL\Row\Entry;
+use Flow\ETL\Row\Reference;
+use Flow\ETL\Row\Schema\Definition;
+use Flow\ETL\Row\Schema\FlowMetadata;
+use Flow\ETL\Row\Schema\Metadata;
+
+/**
+ * @implements Entry<array<array-key, mixed>, array{name: string, map: array<array-key, mixed>, metadata: array{key_type: string, value_type: string}}>
+ */
+final class MapEntry implements \Stringable, Entry
+{
+    use EntryRef;
+
+    /**
+     * @var array{key_type: string, value_type: string}
+     */
+    private array $metadata = [
+        'key_type' => 'null',
+        'value_type' => 'null',
+    ];
+
+    /**
+     * @param array<array-key, mixed> $map
+     *
+     * @throws InvalidArgumentException
+     */
+    public function __construct(
+        private readonly string $name,
+        private readonly array $map
+    ) {
+        if ('' === $name) {
+            throw InvalidArgumentException::because('Entry name cannot be empty');
+        }
+
+        $keysAsString = false;
+        $keysAsInteger = false;
+        $valuesType = 'null';
+
+        $getValueType = function (mixed $value) : string {
+            if (\is_object($value)) {
+                return 'object<' . $value::class . '>';
+            }
+
+            return \gettype($value);
+        };
+
+        foreach ($map as $key => $value) {
+            if (!$keysAsString && \is_string($key)) {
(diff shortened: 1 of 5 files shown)


src/core/etl/src/Flow/ETL/Row/Entry/MapEntry.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public function toString() : string
+    {
+        return \json_encode(['key' => $this->key->toString(), 'value' => $this->value->toString()], \JSON_THROW_ON_ERROR);
+    }
+
+    public function value() : Entry
+    {
+        return $this->value;
```
norberttech: It should be [$this->key->value(), $this->value->value()] - but if uuid/datetime is used, key should be first casted to String. In general you want to have numeric/stringable keys

stloyd: Hmm, isn't just forcing string type for key & value enough?

src/core/etl/src/Flow/ETL/Row/Entry/MapEntry.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return \json_encode(['key' => $this->key->toString(), 'value' => $this->value->toString()], \JSON_THROW_ON_ERROR);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function value() : array
+    {
+        return [(string) $this->key->value(), (string) $this->value->value()];
```
norberttech: you can't cast key->value() to string because then numeric maps will become string maps, casting needs to be done before only for DateTime/Uuid

norberttech:
Hey, everything besides Metadata looks excellent. Let me explain why. 
In strongly typed languages, Maps come with a type for key and value, for example, a map of int to string. 
```
[
    0 => 'a',
    1 => 'b'
]
```

In that case, when writing data to formats like Parquet or Avro, we create something similar to `NestedColumn::map('map', MapKey::int32(), MapValue::string())` . 

With the current shape of Metadata, conversion between Flow schema and Parquet/Avro schema would not be possible because we have no guarantee that it won't be a mixed map type where keys are unique, but sometimes they are integers and sometimes strings. 

In theory, we already have something that could be reused (but it would need to be extracted from the TypeCollection namespace to something more generic). 

- [ScalarType](https://github.com/flow-php/flow/blob/1.x/src/core/etl/src/Flow/ETL/Row/Entry/TypedCollection/ScalarType.php)
- [ObjectType](https://github.com/flow-php/flow/blob/1.x/src/core/etl/src/Flow/ETL/Row/Entry/TypedCollection/ObjectType.php) 

What we could do, is the following. 

- Extract type definitions
- Define on a map entry level what type of the keys/values 
- Validate input $map if keys/values are the expected type 

But it goes even beyond that because, technically speaking, Maps can accept other Maps, Lists, Structures, literally anything as values. 
To make it work, we would need to add some additional types to those that we already have defined (scalar/object), like List (with type of elements), Map (with type of key/value), and maybe Structure (with type for each child). 

I think we should start by defining those generic data types, then implement them in ListEntry (as we already have it and as it's already using it partially), and then move back to this Map.

Issue #715, opened 2023-11-02
Apply limit directly to Extractors


norberttech:
The easiest way to explain it is through example: 

```php
(new Flow())
    ->read(CSV::from(__FLOW_OUTPUT__ . '/dataset.csv'))
    ->limit(100)
    ->write(\Flow\ETL\DSL\To::output())
    ->run();
```

In this case, we want to read only 100 from X rows available in the file. 
Since the limit is now implemented as a pipeline (LimitingPipeline) extractor will still read at least 1000 rows since this is the default value for `$rows_in_batch`. 
We can change this value to 100 however, it requires from the user some understanding of how Extractors are working, which I think is completely unnecessary. 

Instead, limit should be passed directly to the Extractor (we can implement another interface, LimitedExtractor) for example which would let "something" to set that limit so the extractor regardless how it's configured, will not read more rows then in a given limit. 

In theory that's the perfect task for LogicalPlan and Execution Processor but there is a small catch here. 
In the current implementation, LogicalPlan has only access to Pipes (Transformers/Loaders). It does not have access to the Piepline itself, which means it has no access to LimitingPipeline. 

What we need is something between DataFrame and Pipeline, that something is Executor (similar concept exists in Apache Spark).

The role of the Executor would be to create a LogicalPlan, process it using Execution Processor, and execute the pipeline. 
Thanks to that, we are going to get direct access to Pipeline (which might hold other pipelines under itself), which would give us more flexible way of analyzing the entire execution chain and adjusting it (optimizing/validating/changing) before it gets executed. 

Now going back to applying the limit on the Extractor. 

So Processor should check if there are any transformations in the pipeline before DataFrame::limit() is used, whenever it wont find anything that might change the number of rows (like for example, expand), it should apply the limit directly on the Executor removing LimitingPipeline. Let's look at some examples: 

```php
(new Flow())
    ->read(CSV::from(__FLOW_OUTPUT__ . '/dataset.csv'))
    ->select("id", "name")
    ->limit(100)
    ->write(\Flow\ETL\DSL\To::output())
    ->run();
```

In this example, there is nothing before the limit that might change the number of results, limit should be applied directly on the Extractor. 

```php
(new Flow())
    ->read(CSV::from(__FLOW_OUTPUT__ . '/dataset.csv'))
    ->withEntry("item", ref("items")->expand()) 
    ->limit(100)
    ->write(\Flow\ETL\DSL\To::output())
    ->run();
```

[...]

PR #716, opened 2023-11-03
Changed Loader Closure definition

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>batchSize argument to DataFrame::collect method</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Closure was moved to Loader namespace as it applies to Loaders, not Pipelines</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Rows from Closure::closure method</li>
    <li>BufferLoader</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

Ref: #715 

<h2>Description</h2>

BufferLoader was implemented at very early stage of ETL, it's goal was to reduce number of IO operations when writing data. From now the same behavior can be achieved by setting a batchSize on `DataFrame::collect()` method. Before collect was collecting everything into memory, and it still does that when batchSize is not specified. 
When batchSize is defined, CollectingPipeline will keep merging rows until they reach batchSize and then it will yield them. 

Why is this important? 

Thanks to the collect method, users have control over how many Rows at once are passed to the Loader, which can help in achieving better performance or displaying more readable output. 

We can also remove batchSize from the Loaders directly, as they should not be responsible for dividing received Rows into chunks, but it's a thing for the next PR.


src/adapter/etl-adapter-avro/src/Flow/ETL/Adapter/Avro/FlixTech/AvroLoader.php
```diff
@@ -49,7 +49,7 @@ public function __unserialize(array $data) : void
         $this->writer = null;
     }
 
-    public function closure(Rows $rows, FlowContext $context) : void
+    public function closure(FlowContext $context) : void
```
stloyd, line 52: As this is public I think it also requires note in upgrade file.

src/core/etl/src/Flow/ETL/DSL/To.php
```diff
@@ -18,14 +18,6 @@
  */
 class To
 {
-    /**
-     * @param int<1, max> $bufferSize
-     */
-    final public static function buffer(Loader $overflowLoader, int $bufferSize) : Loader
```
stloyd, line 24: This most likely also requires note in upgrade file.

stloyd (commented):
I like the changes, just small notes that seem to be missing from the upgrade file

src/core/etl/src/Flow/ETL/DSL/To.php
norberttech: it's already there

norberttech:
Two more things can be implemented after this is merged: 

#717 
#718

src/core/etl/src/Flow/ETL/DSL/To.php
stloyd: It says about `BufferLoader`, but not about the `To::buffer()` method, which is related but still different place.

norberttech: it's the same thing, To::buffer is only a DSL function to simplify initialization of BufferLoader but feel free to add it into the upgrade notes

Issue #722, opened 2023-11-03
Snappy & Parquet tests can throw warnings randomly


stloyd:
As noticed locally, we can see it also in CI:
- https://github.com/flow-php/flow/actions/runs/6766449539/job/18387564614
- https://github.com/flow-php/flow/actions/runs/6766449539/job/18387564713
- https://github.com/flow-php/flow/actions/runs/6766449539/job/18387564787

Randomly happening test warnings:
```console
2 tests triggered 2 PHP warnings:

1) /Users/stloyd/Documents/flow/src/lib/snappy/src/Flow/Snappy/SnappyCompressor.php:221
Undefined array key 2693

Triggered by:

* Flow\Parquet\Tests\Integration\IO\MapsWritingTest::test_writing_map_of_int_int
  /Users/stloyd/Documents/flow/src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/MapsWritingTest.php:17

2) /Users/stloyd/Documents/flow/src/lib/parquet/src/Flow/Parquet/BinaryReader/Bytes.php:50
Undefined array key 0

Triggered by:

* Flow\ETL\Adapter\Parquet\Tests\Integration\ParquetTest::test_writing_with_partitioning
  /Users/stloyd/Documents/flow/src/adapter/etl-adapter-parquet/tests/Flow/ETL/Adapter/Parquet/Tests/Integration/ParquetTest.php:46

stloyd:
Closing in favor of #727

stloyd:
Re-opening the issue as #727 was not enough to cover the bug, see updated description.

PR #725, opened 2023-11-04
Convert PHP warnings to test failures

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Convert PHP warnings to test failures</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Related to #722

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -170,6 +170,9 @@
         "test": [
             "tools/phpunit/vendor/bin/phpunit"
         ],
+        "test:no-coverage": [
+            "tools/phpunit/vendor/bin/phpunit --no-coverage"
+        ],
         "test:benchmark": [
             "@test:benchmark:building_blocks",
             "@test:benchmark:extractor",
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,6 +7,8 @@
         cacheResultFile="./var/phpunit/phpunit.cache"
         displayDetailsOnTestsThatTriggerWarnings="true"
         displayDetailsOnTestsThatTriggerErrors="true"
+        failOnNotice="true"
+        failOnWarning="true"
         executionOrder="random"
 >
   <coverage cacheDirectory="./var/phpunit/">


composer.json
```diff
@@ -168,7 +168,7 @@
             "@test:mutation"
         ],
         "test": [
-            "tools/phpunit/vendor/bin/phpunit"
+            "tools/phpunit/vendor/bin/phpunit --no-coverage"
```
norberttech, line 174: This should be passed directly at GitHub actions, I do use coverage from `composer test` locally and by making this change here, it will be no longer available.

stloyd: What about adding: `test:coverage` aside of this?

norberttech: Well technically speaking one of the steps before opening PR should be to check coverage which means that `composer build` will no longer generate coverage, if anything, it could be `test:no-coverage`. 
What's the use case for executing tests without coverage?

stloyd: I don't install xdebug/pcov by default locally as it's slower, thus I always get PHPUnit warning about missing coverage driver.

Anyway reverted & added a new call for usage without coverage that works for me.

norberttech: Hmm right, I think we also need to look into displaying coverage for the new code through GitHub actions, it might be easy to miss that something important wasn't covered when using this option.

norberttech:
This can to be merged only after #727 is fixed

norberttech:
#731 is going to make this merge possible, but please wait a bit, I was thinking about it, and I'm not sure if I this is going to bring any value. If the goal is to avoid merging code with PHPUnit warnings, we might want to start from adding `--fail-on-warning` on the CI.

stloyd:
Having this only on CI means that the developer can be not aware of the problem (it will not be marked locally as a failure), but it may happen on CI. Another problem is that if it's more random, it may happen on CI in some cases, but most likely will be ignored locally as "tests passed", especially if a developer will not run all tests like `composer test` but only those related to code he/she works on.

norberttech:
> Having this only on CI means that the developer can be not aware of the problem (it will not be marked locally as a failure), but it may happen on CI.

That's pretty much all I care about. 

> Another problem is that if it's more random, it may happen on CI in some cases, but most likely will be ignored locally as "tests passed", especially if a developer will not run all tests like composer test but only those related to code he/she works on.

The example of our most recent warning showed that fixing warnings is not so simple. 
It might require deeper debugging and understanding of the root cause of the problem. I'm better (at least for now) doing it by myself since it's not happening every day. It's usually caused by some edge case. 
And if the warning was introduced in the just opened PR (which still is not something that happens with every PR), dev should be able to identify it quickly and fix it.

I understand the reasoning behind that change, in theory it makes sense, but I tried that before, and it caused just some frustration without any additional value given. Like when one of the dependencies will start throwing deprecation warnings.

I'm closing this for now, but we should keep monitoring how frequently warnings are appearing, and I still think we can put `--fail-on-warning` as a first step, not only to prevent merging something that is throwing warnings, but also to detect issues with tests using random data. However, I would do it only for locked dependencies, highest/lowest should still pass. regardless of the number of warnings.

PR #728, opened 2023-11-04
Extract entry types into new namespace for further re-usage

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Extract entry types into new namespace for further re-usage</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/core/etl/src/Flow/ETL/Row/Entry/Type/NativeObject.php
```diff
@@ -2,11 +2,11 @@
 
 declare(strict_types=1);
 
-namespace Flow\ETL\Row\Entry\TypedCollection;
+namespace Flow\ETL\Row\Entry\Type;
 
 use Flow\ETL\Exception\InvalidArgumentException;
 
-final class ObjectType implements Type
+final class NativeObject implements Type
```
norberttech: yeah, the reason why I kept ObjectType was that Object is a reserved keyword which IMO sounds better than NativeObject - that's why I also left ScalarType. 

`ObjectType::of(\DateTimeImmutableClass)` also looks more descriptive than `ObjectType::fromString()` - maybe `ObjectType::fromClass()` would be better.

src/core/etl/src/Flow/ETL/Row/Entry/Type/NativeObject.php
```diff
@@ -19,35 +19,23 @@ public function __construct(public readonly string $class)
...
+        return new self($value);
     }
 
     public function isEqual(Type $type) : bool
     {
         return $type instanceof self && $type->class === $this->class;
     }
 
-    public function isValid(array $collection) : bool
+    public function isValid(mixed $value) : bool
```
norberttech: 👍

src/core/etl/src/Flow/ETL/Row/Entry/Type/Type.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Row\Entry\Type;
```
norberttech: I'm not 100% sure if that namespace is the correct one. I think I would take it even higher, to `\Flow\ETL\PHP\Type` maybe since those objects are representing PHP Types

src/core/etl/src/Flow/ETL/Row/Entry/Type/Uuid.php
```diff
@@ -5,7 +5,7 @@
 use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\Exception\RuntimeException;
 
-final class Uuid implements \Stringable
+final class Uuid implements \Stringable, Type
```
norberttech: hmm that is an interesting approach, I thought this was going to be handled by ObjectType when used in Maps/Lists?

src/core/etl/src/Flow/ETL/Row/Entry/Type/NativeObject.php
stloyd: TBH. I have hard to swallow problem with the `ObjectType` itself, I cannot find a real case for its usage in my head, as from my perspective rather converter(?) alike, than real type, especially when looking at the parquet [definition of logical types](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists) where value is "just" converted to binary, where here it would (probably) string.

norberttech: Map of ints to \DateTimeImmutable is an example of where ObjectType would be an type of Map Element, same with Lists, for example List of Uuid's

src/core/etl/src/Flow/ETL/Row/Entry/Type/Type.php
stloyd: Moved there.

src/core/etl/src/Flow/ETL/Row/Entry/Type/Uuid.php
stloyd: Reverted.

src/core/etl/src/Flow/ETL/Row/Entry/Type/NativeObject.php
stloyd: Restored old name.

norberttech:
Awesome, this is a great starting point for Maps and improving Lists, but I think we might also want to use those types in Expressions later.

PR #729, opened 2023-11-04
Replace LimitingPipeline with LimitTransformer

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>LimitTransformer</li>
    <li>Limit directly to Extractors</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>LimitPipeline</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

Refs: #715 
<h2>Description</h2>

In order to analyze the Execution Plan we need to iterate over Pipeline elements, since Limit was implemented as Pipeline itself it was pretty much impossible to understand if/when limit was applied to the pipeline. 
This approach changes that, instead of using LimitingPipeline, there is a new Transformer that will throw LimitReached exception which should be recognized to send Signal::STOP to generator. 
All generators are now expecting Signals, once Singal::STOP is received, generators are closed. 

This PR comes also with adding Limit directly to Extractors, it's for the performance auto optimization that will happen in the next PR. 
Once limit is applied to the Extractor, it will throw only specific number of rows, some extractors like Parquet will use that limit to read only specific number of values instead of reading the entire page of data.


src/adapter/etl-adapter-avro/src/Flow/ETL/Adapter/Avro/FlixTech/AvroExtractor.php
```diff
@@ -41,7 +47,14 @@ public function extract(FlowContext $context) : \Generator
...
                 }
 
-                yield array_to_rows($row, $context->entryFactory());
+                $signal = yield array_to_rows($row, $context->entryFactory());
+                $this->countRow();
+
+                if ($signal === Signal::STOP || $this->reachedLimit()) {
+                    $context->streams()->close($this->path);
+
+                    return;
```
stloyd, line 56: in all file-based ones it could be replaced with:
```suggestion
                    break 2;
```
along with that closing the stream could be removed as would be done at the end of that method.

src/core/etl/src/Flow/ETL/Extractor/Signal.php
```diff
@@ -0,0 +1,8 @@
+<?php declare(strict_types=1);
+
+namespace Flow\ETL\Extractor;
+
+enum Signal
+{
+    case STOP;
```
stloyd, line 7: What about `STOP_PROCESSING`? Cause stop is quite a generic name

stloyd (approved):
I like this approach 😊

src/adapter/etl-adapter-avro/src/Flow/ETL/Adapter/Avro/FlixTech/AvroExtractor.php
norberttech: I started from this approach but return is easier to understand and less problematic in case of additional logic after the loop. Return is more readable at least for me

src/core/etl/src/Flow/ETL/Extractor/Signal.php
norberttech: Extractors are not really processing, they are extracting, but I didn't want yet to implement something super specific, I want to wait and see how this will evolve, the moment STOP will become too enigmatic, then we might need to think about splitting it

PR #730, opened 2023-11-04
Pipeline Optimization - LimitOptimizer

### LINKED ISSUE #715: Apply limit directly to Extractors

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Pipeline Optimizer</li>
    <li>LimitOptimization</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Renamed LogicalPlan into ExecutionPlan</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

Closes: #715 

<h2>Description</h2>

I was first trying to extend ExecutionPlan processors, however achieving what was needed without access to the Pipeline itself was extremely complicated, I ended up adding Locks to Transformers in order to avoid removing them from the pipeline in certain situations. 
After a few attempts, I started once again from scratch, this time focusing on optimizing the pipeline not before execution but when the element is supposed to be added to the pipeline. 

That's how I ended up creating a new mechanism in the ETL called `Optimizer`. 
The role of the optimizer is to analyze the existing pipeline/extractor and decide what's the most optimal way of adding a given element to the pipeline. 

Currently, there is only one Optimization that can be applied, `LimitOptimization` but we can easily expand this mechanism in order to: 

- reduce the number of transformers, merging the same transformations together (like keep/drop entries) 
- apply selects/filters directly at extractors (when possible) 
- remove redundant transformations 

It's best to show it on example: 

```php
<?php

use Flow\ETL\DSL\Parquet;
use Flow\ETL\DSL\To;
use Flow\ETL\Flow;

include __DIR__ . '/../vendor/autoload.php';
include __DIR__ . '/../tools/blackfire/vendor/autoload.php';

$start = hrtime(true);
(new Flow())
    ->read(Parquet::from(__DIR__ . '/1_mln_rows.parquet',))
    ->limit(10)
    ->write(To::output())
    ->count();
$end = hrtime(true);
echo 'Time: ' . ($end - $start) / 1e+9 . 's' . PHP_EOL;
```

Without Optimizations total time of reading 10 rows from 1mln rows parquet file was something around 11 sec, after optimization is applied, that time drops below 1sec.


src/core/etl/src/Flow/ETL/Pipeline.php
```diff
@@ -22,10 +24,14 @@ public function has(string $transformerClass) : bool;
...
+
     /**
      * @return \Generator<Rows>
      */
     public function process(FlowContext $context) : \Generator;
 
-    public function source(Extractor $extractor) : self;
+    public function setSource(Extractor $extractor) : self;
+
+    public function source() : Extractor;
```
norberttech, line 36: I had to expose the pipeline data source in order to be able to access it at the Optimizer level.

src/core/etl/src/Flow/ETL/Pipeline/Execution/ExecutionPlan.php
```diff
@@ -8,7 +8,7 @@
 use Flow\ETL\Pipeline\Execution\Plan\FilesystemOperations;
 use Flow\ETL\Pipeline\Pipes;
 
-final class LogicalPlan
+final class ExecutionPlan
```
norberttech, line 11: Technically speaking, looking at the current implementation, it's more a Physical than a Logical plan, but for now, lets call it an Execution Plan which IMO it's a most self-descriptive name for what this class is really responsible for.

norberttech:
Technically speaking, it should be possible to move the current Execution Plan processing logic into Optimizer, however, for now I would like to keep both mechanisms. There is a very good chance that Execution Plan processor will evolve into `Executor`, which I still think it's missing. 
For example, now, asynchronous processing is possible through the LocalSocketPipeline (async) pipeline. Sync/Async could be moved to the Executor.

Issue #753, opened 2023-11-06
Extract logic for native & logical types guess of from `NativeEntryFactory`


stloyd:
Our `NativeEntryFactory` grows too big and handles too much logic, we should extract types guessing into a new factory.

PR #772, opened 2023-11-08
Add `TypeFactory`, `ArrayType`, `NullType`, `ResourceType` & `CallableType`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `TypeFactory`, `ArrayType` & `NullType`, `ResourceType`, `CallableType`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Refs #753

### DIFF
--- a/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
+++ b/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
@@ -6,13 +6,14 @@ namespace Flow\ETL\PHP\Type;
 
 use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\PHP\Type\Native\ArrayType;
+use Flow\ETL\PHP\Type\Native\NullType;
 use Flow\ETL\PHP\Type\Native\ScalarType;
 
 final class ArrayContentDetector
 {
-    private readonly Type $firstKeyType;
+    private readonly ?Type $firstKeyType;
 
-    private readonly Type $firstValueType;
+    private readonly ?Type $firstValueType;
 
     public function __construct(private readonly Types $uniqueKeysType, private readonly Types $uniqueValuesType)
     {
@@ -20,41 +21,41 @@ final class ArrayContentDetector
         $this->firstValueType = $this->uniqueValuesType->first();
     }
 
-    public function firstKeyType() : ScalarType
+    public function firstKeyType() : ?ScalarType
     {
-        if (!$this->firstKeyType instanceof ScalarType) {
+        if (null !== $this->firstKeyType && !$this->firstKeyType instanceof ScalarType) {
             throw InvalidArgumentException::because('First unique key type must be of ScalarType, given: ' . $this->firstKeyType::class);
         }
 
         return $this->firstKeyType;
     }
 
-    public function firstValueType() : Type
+    public function firstValueType() : ?Type
     {
         return $this->firstValueType;
     }
 
     public function isList() : bool
     {
-        if (!$this->firstKeyType()->isInteger()) {
+        if (!$this->firstKeyType()?->isInteger()) {
             return false;
         }
 
-        return 1 === $this->uniqueValuesType->filter(fn (Type $type) : bool => !($type instanceof ArrayType && $type->empty()))->count();
+        return 1 === $this->uniqueValuesType->without(new ArrayType(true), new NullType())->count();
     }
 
     public function isMap() : bool
     {
-        if (1 === $this->uniqueValuesType->count()) {
+        if (1 === $this->uniqueValuesType->without(new ArrayType(true), new NullType())->count()) {
             if ($this->isList()) {
                 return false;
             }
 
-            if (!($this->firstKeyType()->isString() || $this->firstKeyType()->isInteger())) {
+            if (!($this->firstKeyType()?->isString() || $this->firstKeyType()?->isInteger())) {
                 return false;
             }
 
-            return 1 === $this->uniqueKeysType->filter(fn (Type $type) : bool => !($type instanceof ArrayType && $type->empty()))->count();
+            return 1 === $this->uniqueKeysType->count();
         }
 
         return false;
@@ -66,6 +67,8 @@ final class ArrayContentDetector
             return false;
         }
 
-        return $this->firstKeyType()->isString() && 1 === $this->uniqueKeysType->count();
+        return $this->firstKeyType()?->isString()
+            && 1 === $this->uniqueKeysType->count()
+            && 0 !== $this->uniqueValuesType->without(new ArrayType(true), new NullType())->count();
     }
 }
--- a/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
+++ b/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
@@ -1,51 +1,31 @@
 <?php
+
 declare(strict_types=1);
 
 namespace Flow\ETL\PHP\Type;
 
 use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\PHP\Type\Native\ArrayType;
-use Flow\ETL\PHP\Type\Native\NullType;
 use Flow\ETL\PHP\Type\Native\ScalarType;
 
 final class ArrayContentDetector
 {
-    private readonly ScalarType $firstKeyType;
+    private readonly Type $firstKeyType;
 
     private readonly Type $firstValueType;
 
-    private readonly array $uniqueKeysType;
-
-    private readonly array $uniqueValuesType;
-
-    public function __construct(
-        array $keysTypes,
-        array $valuesTypes,
-    ) {
-        if (0 === \count($keysTypes)) {
-            throw new InvalidArgumentException('Key type list cannot be empty');
-        }
-
-        if (0 === \count($valuesTypes)) {
-            throw new InvalidArgumentException('Value type list cannot be empty');
-        }
-
-        $this->uniqueKeysType = \array_unique(\array_map(fn (Type $type) : string => \serialize($type), $keysTypes));
-        $this->uniqueValuesType = \array_unique(
-            \array_map(fn (Type $type) : string => \serialize($type), \array_filter($valuesTypes, function (Type $type) : bool {
-                if ($type instanceof NullType) {
-                    return false;
-                }
-
-                return !($type instanceof ArrayType && $type->empty());
-            }))
-        );
-        $this->firstKeyType = \unserialize($this->uniqueKeysType[0]);
-        $this->firstValueType = \unserialize($this->uniqueValuesType[0]);
+    public function __construct(private readonly Types $uniqueKeysType, private readonly Types $uniqueValuesType)
+    {
+        $this->firstKeyType = $this->uniqueKeysType->first();
+        $this->firstValueType = $this->uniqueValuesType->first();
     }
 
     public function firstKeyType() : ScalarType
     {
+        if (!$this->firstKeyType instanceof ScalarType) {
+            throw InvalidArgumentException::because('First unique key type must be of ScalarType, given: ' . $this->firstKeyType::class);
+        }
+
         return $this->firstKeyType;
     }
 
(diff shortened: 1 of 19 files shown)


src/core/etl/src/Flow/ETL/PHP/Type/ArrayTypeDetector.php
```diff
@@ -0,0 +1,81 @@
...
+                }
+
+                return !($type instanceof ArrayType && $type->empty());
+            }))
+        );
+    }
+
+    public function firstKeyType() : ScalarType
+    {
+        return \unserialize($this->uniqueKeysType[0]);
```
stloyd: @norberttech The only thing I'm considering is to save this as `readonly` variables in the constructor.

src/core/etl/src/Flow/ETL/PHP/Type/Native/ArrayType.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function isValid(mixed $value) : bool
+    {
+        return \is_array($value);
+    }
+
+    public function toString() : string
+    {
+        return 'array<mixed>';
```
norberttech, line 35: when empty I think it should be something like `array-empty` or maybe `array<empty, empty>`

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
```diff
@@ -0,0 +1,72 @@
...
+use Flow\ETL\PHP\Type\Logical\StructureType;
+use Flow\ETL\PHP\Type\Native\ArrayType;
+use Flow\ETL\PHP\Type\Native\NullType;
+use Flow\ETL\PHP\Type\Native\ObjectType;
+use Flow\ETL\PHP\Type\Native\ScalarType;
+
+final class TypeFactory
+{
+    public function getType(mixed $value) : Type
+    {
```
norberttech, line 23: I think you should check if it's not a resource or callable as well

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        if (\is_object($value)) {
+            return ObjectType::of($value::class);
+        }
+
+        if (\is_array($value)) {
+            if ([] === \array_filter($value, fn ($value) : bool => null !== $value)) {
+                return new ArrayType(true);
+            }
+
+            $detector = new ArrayTypeDetector(
```
norberttech: `ArrayContentDetector` should a bit more descriptive, now when I think about it

src/core/etl/src/Flow/ETL/PHP/Type/ArrayTypeDetector.php
norberttech: yeah, readonly sounds good, this is not something that should changed after object is created

src/core/etl/src/Flow/ETL/PHP/Type/ArrayTypeDetector.php
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        $this->uniqueKeysType = \array_unique(\array_map(fn (Type $type) : string => \serialize($type), $keysTypes));
+        $this->uniqueValuesType = \array_unique(
+            \array_map(fn (Type $type) : string => \serialize($type), \array_filter($valuesTypes, function (Type $type) : bool {
+                if ($type instanceof NullType) {
+                    return false;
+                }
+
+                return !($type instanceof ArrayType && $type->empty());
```
norberttech: I don't think it should be filtered out here, I would move that filter logic to isMap/isList directly

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
stloyd: You mean by explicit? If all checks fail, it will end-up with an exception at the end, so IMO it's covered already (ofc can add a test for it).

src/core/etl/src/Flow/ETL/PHP/Type/ArrayTypeDetector.php
norberttech: Another thing, that would make this class even easier to read would be to create a class called `Types` that would be a collection of Types with all handy methods on it, like filtering nulls etc

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
norberttech: nah, I mean that we should have a type for Resource and Callables since array of resources or array of callables or list/map of resource/callables is something that should be supported.

src/core/etl/src/Flow/ETL/PHP/Type/ArrayTypeDetector.php
stloyd: One thing here is that we have this filter before serialize, later we would need to unserialize to get the types.

norberttech: Yeah but without that, we won't be able to properly detect if lists/maps are nullable

src/core/etl/src/Flow/ETL/PHP/Type/Native/ArrayType.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public function empty() : bool
+    {
+        return $this->empty;
+    }
+
+    public function isEqual(Type $type) : bool
+    {
+        return $type instanceof self;
```
norberttech: You should also compare `$this->empty`

src/core/etl/src/Flow/ETL/PHP/Type/Types.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function count() : int
+    {
+        return \count($this->types);
+    }
+
+    /**
+     * @param callable(Type) : bool $callable
+     */
+    public function filter(callable $callable) : self
```
norberttech: I think I would replace it with something less generic, like `Types::without(ArrayType::empty(), new NullType())`

src/core/etl/src/Flow/ETL/PHP/Type/Types.php
```diff
@@ -0,0 +1,58 @@
...
+    private readonly array $types;
+
+    private function __construct(Type ...$types)
+    {
+        $this->types = \array_map(
+            fn (string $type) : Type => \unserialize($type),
+            \array_unique(
+                \array_map(
+                    fn (Type $type) : string => \serialize($type),
+                    \array_filter($types, fn (Type $type) : bool => !$type instanceof NullType)
```
norberttech: This should be requested intentionally, not filtered out by default, because we are losing knowledge about nulls in array

src/core/etl/src/Flow/ETL/PHP/Type/Types.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param callable(Type) : bool $callable
+     */
+    public function filter(callable $callable) : self
+    {
+        return new self(...\array_filter($this->types, $callable));
+    }
+
+    public function first() : Type
+    {
+        return $this->types[0];
```
norberttech: In this case you should verify in constructor if there is at least one, or return here ?Type on empty Types. Especially when this collection can be filtered.

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
```diff
@@ -0,0 +1,72 @@
...
+        if (\is_scalar($value)) {
+            return ScalarType::fromString(\gettype($value));
+        }
+
+        if (\is_object($value)) {
+            return ObjectType::of($value::class);
+        }
+
+        if (\is_array($value)) {
+            if ([] === \array_filter($value, fn ($value) : bool => null !== $value)) {
```
norberttech: this logic should be covered by detector that should not return true on isMap/isList/isStruct, when array is empty or its an array of nulls.

src/core/etl/src/Flow/ETL/PHP/Type/Types.php
```diff
@@ -0,0 +1,58 @@
...
+    public static function create(Type ...$types) : self
+    {
+        if (0 === \count($types)) {
+            throw new InvalidArgumentException('Type list cannot be empty');
+        }
+
+        return new self(...$types);
+    }
+
+    public function all() : array
```
norberttech: hmm, how is it possible that static analysis is not complaining about the missing type of $this->types?

src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
```diff
@@ -0,0 +1,74 @@
...
+final class ArrayContentDetector
+{
+    private readonly ?Type $firstKeyType;
+
+    private readonly ?Type $firstValueType;
+
+    public function __construct(private readonly Types $uniqueKeysType, private readonly Types $uniqueValuesType)
+    {
+        $this->firstKeyType = $this->uniqueKeysType->first();
+        $this->firstValueType = $this->uniqueValuesType->first();
```
norberttech: You can just cache it in the Types, and here just return `$this->uniqueValuesType->first()` when it's needed

src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
```diff
@@ -0,0 +1,74 @@
...
+        return $this->firstValueType;
+    }
+
+    public function isList() : bool
+    {
+        if (!$this->firstKeyType()?->isInteger()) {
+            return false;
+        }
+
+        return 1 === $this->uniqueValuesType->without(new ArrayType(true), new NullType())->count();
```
norberttech: It's a small change but I would replace `new ArrayType(true)` with `ArrayType::empty()` to make it more descriptive

src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function isMap() : bool
+    {
+        if (1 === $this->uniqueValuesType->without(new ArrayType(true), new NullType())->count()) {
+            if ($this->isList()) {
+                return false;
+            }
+
+            if (!($this->firstKeyType()?->isString() || $this->firstKeyType()?->isInteger())) {
```
norberttech: those two methods could be replaced with `isValidArrayKey()`

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
```diff
@@ -0,0 +1,72 @@
...
+                $elements = [];
+
+                foreach ($value as $key => $item) {
+                    $elements[] = new StructureElement($key, $this->getType($item));
+                }
+
+                return new StructureType(...$elements);
+            }
+
+            return new ArrayType([] === \array_filter($value, fn ($value) : bool => null !== $value));
```
norberttech, line 67: `\array_filter($value, fn ($value) : bool => null !== $value)` not sure if I understand this part, could you explain it to me?

stloyd: Without such filter this example will fail:
```php
        yield 'list of lists with array of nulls' => [
            [
                [
                    1, 2, 3,
                ],
                [
                    null,
                ],
                [
                    4, 5, 6,
                ],
            ],
            ListType::class,
            'list<list<integer>>',
        ];

src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
stloyd: You mean `private` one in detector?

norberttech: `$this->firstKeyType?->isValidArrayKey()`

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
norberttech: ah, I see it now, it's a recursive execution 👍

norberttech (approved):

PR #778, opened 2023-11-11
Fixed Dremel to properly shred/assemble nested structures with nullable elements

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Dremel to properly shred/assemble nested structures with nullable elements</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Dremel algorithms are no longer working as Generators</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

While working on Dremel algorithms implementation I was mostly focused on non nullable nested structures, that's how this bug was missed in the first implementation. 
After this fix, it will not only properly handle maps/lists with nulls but also empty maps/lists. 
Dremel is no longer working as a generator because all data needs to sit in the memory anyway and since we are operating at small integers, this was bringing very little value compared to complexity of debugging it.

### DIFF
--- a/src/lib/dremel/tests/Flow/Dremel/Tests/Unit/DremelShredTest.php
+++ b/src/lib/dremel/tests/Flow/Dremel/Tests/Unit/DremelShredTest.php
@@ -4,7 +4,6 @@ namespace Flow\Dremel\Tests\Unit;
 
 use function Flow\Parquet\array_flatten;
 use Flow\Dremel\Dremel;
-use Flow\Dremel\Exception\RuntimeException;
 use PHPUnit\Framework\TestCase;
 
 final class DremelShredTest extends TestCase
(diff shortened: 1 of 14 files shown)


src/lib/dremel/tests/Flow/Dremel/Tests/Unit/DremelShredTest.php
```diff
@@ -103,25 +121,4 @@ public function test_dremel_shred_on_repeated_columns() : void
             ]
         );
     }
-
-    public function test_dremel_with_rows_with_combined_scalar_values_and_array_values() : void
```
norberttech, line 107: This constraint was incorrect since there is a scenario where arrays also contain nulls, not only scalar values.

Issue #775, opened 2023-11-09
Refactor `StructureEntry` to use a new logical `StructureType`


stloyd:
Right now `StructureEntry` uses an array of entries as a value for structure, it's not fully correct, and we should change it to use the same as we have for `MapEntry` (ref #684) along with a new logical `StructureType`.

PR #780, opened 2023-11-11
Rework `StructureEntry` & add `MapEntry`

### LINKED ISSUE #684: MapEntry - new entry type

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `MapEntry`</li>
    <li>Add `EnumType`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Rework `StructureEntry` to use types</li>
    <li>Rework `NativeEntryFactory` to use types</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes #684, #775


src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
```diff
@@ -40,26 +40,9 @@ public function isList() : bool
         return 1 === $this->uniqueValuesType->without(ArrayType::empty(), new NullType())->count();
     }
 
-    public function isMap() : bool
```
norberttech: I think I would keep this method for now, we are not going to use it here, but it could be used by a MapEntry to validate if the given input types are valid for the map

src/core/etl/src/Flow/ETL/DSL/Entry.php
```diff
@@ -276,27 +262,32 @@ final public static function string_upper(string $name, string $value) : RowEntr
         return RowEntry\StringEntry::uppercase($name, $value);
     }
 
-    final public static function struct(string $name, RowEntry ...$entries) : RowEntry
```
norberttech, line 255: I would like to keep it as an alist for `structure()`

src/core/etl/src/Flow/ETL/DSL/Entry.php
```diff
@@ -276,27 +262,32 @@ final public static function string_upper(string $name, string $value) : RowEntr
...
-    {
-        return self::structure($name, ...$entries);
-    }
-
     /**
      * @throws InvalidArgumentException
      *
      * @return RowEntry\StructureEntry
      */
     final public static function structure(string $name, RowEntry ...$entries) : RowEntry
```
norberttech: This approach is no longer valid since the structure was refactored. What it should take instead StructType and array of values

src/core/etl/src/Flow/ETL/PHP/Type/Native/ScalarType.php
```diff
@@ -118,11 +124,6 @@ public function isValid(mixed $value) : bool
         return true;
     }
 
-    public function isValidArrayKey() : bool
```
norberttech: I would also keep this method present as it will be relevant for maps

src/core/etl/src/Flow/ETL/PHP/Type/Native/ObjectType.php
```diff
@@ -25,9 +25,9 @@ public function __construct(public readonly string $class, private readonly bool
     /**
      * @param class-string $class
      */
-    public static function of(string $class) : self
+    public static function of(string $class, bool $nullable = false) : self
```
norberttech, line 33: since we now have a proper ObjectType, ObjectEntry should keep that type internally and pass it to Schema Definition so we can easier validate schema against ObjectEntry just by comparing ObjectTypes.

src/core/etl/tests/Flow/ETL/Tests/Unit/Row/Schema/Formatter/ASCIISchemaFormatterTest.php
```diff
@@ -34,12 +43,12 @@ public function test_format_nested_schema() : void
             <<<SCHEMA
 schema
 |-- user: Flow\ETL\Row\Entry\StructureEntry (nullable = false)
-|    |-- name: [Flow\ETL\Row\Entry\StringEntry, Flow\ETL\Row\Entry\NullEntry] (nullable = true)
-|    |-- age: Flow\ETL\Row\Entry\IntegerEntry (nullable = false)
-|    |-- address: Flow\ETL\Row\Entry\StructureEntry (nullable = false)
-|        |-- street: [Flow\ETL\Row\Entry\StringEntry, Flow\ETL\Row\Entry\NullEntry] (nullable = true)
-|        |-- city: [Flow\ETL\Row\Entry\StringEntry, Flow\ETL\Row\Entry\NullEntry] (nullable = true)
-|        |-- country: [Flow\ETL\Row\Entry\StringEntry, Flow\ETL\Row\Entry\NullEntry] (nullable = true)
+|    |-- name: ?string (nullable = true)
```
norberttech: Since we can't use FQCN of Entry, schema should be modified to follow the same convention. To make it easier, we can additional method to `Entry` interface, `Entry::phpType() : Type` and then just use toString() on that type which should be aligned also with psalm/phpstan types, maybe except list/map/structure but those are anyway logical types.

src/core/etl/tests/Flow/ETL/Tests/Unit/Transformer/EntryExpressionEvalTransformerTest.php
```diff
@@ -94,7 +94,7 @@ public function test_xml_xpath_expression_when_there_is_more_than_one_node_under
         $xpath = new \DOMXPath($document);
 
         $this->assertEquals(
-            Entry::list_of_objects('xpath', \DOMNode::class, [
+            Entry::list_of_objects('xpath', \DOMElement::class, [
```
norberttech, line 97: hmm why?

src/core/etl/src/Flow/ETL/Row/Entry/StructureEntry.php
```diff
@@ -6,49 +6,47 @@
...
+        if (0 === \count($value)) {
             throw InvalidArgumentException::because('Structure must have at least one entry, ' . $name . ' got none.');
         }
 
-        $entryNames = \array_map(static fn (Entry $entry) => $entry->name(), $entries);
-
-        if (\count(\array_unique($entryNames)) !== \count($entryNames)) {
-            throw InvalidArgumentException::because('Each entry name in structure must be unique, given: ' . \implode(', ', $entryNames));
+        if (!$type->isValid($value)) {
+            throw InvalidArgumentException::because('Expected list of ' . $type->toString() . ' got different types.');
```
norberttech: Expected `structure` not `list`. Is StructureType also checking if there are no additional/missing fields in the values?

src/core/etl/src/Flow/ETL/DSL/Entry.php
```diff
@@ -58,21 +62,11 @@ final public static function collection(string $name, Entries ...$entries) : Row
...
     {
         return new RowEntry\DateTimeEntry($name, $value);
     }
 
-    /**
-     * @throws InvalidArgumentException
-     *
-     * @return RowEntry\DateTimeEntry
-     */
-    final public static function datetime_string(string $name, string $value) : RowEntry
```
norberttech: all of the removed DSL methods should be mentioned in the UPGRADE.md

src/core/etl/tests/Flow/ETL/Tests/Unit/Transformer/EntryExpressionEvalTransformerTest.php
stloyd: Cause: `$xpath->query('/root/foo')->item(0)` returns `DOMElement` not just `DOMNode`:
```console
object(DOMElement)#409 (23) {
  ["schemaTypeInfo"]=>
  NULL
  ["tagName"]=>
  string(3) "foo"
  ["firstElementChild"]=>
  NULL
  ["lastElementChild"]=>
  NULL
  ["childElementCount"]=>
  int(0)
  ["previousElementSibling"]=>
  NULL
  ["nextElementSibling"]=>
  string(22) "(object value omitted)"
  ["nodeName"]=>
  string(3) "foo"
  ["nodeValue"]=>
  string(3) "bar"
  ["nodeType"]=>
  int(1)
  ["parentNode"]=>
  string(22) "(object value omitted)"
  ["childNodes"]=>
  string(22) "(object value omitted)"
  ["firstChild"]=>
  string(22) "(object value omitted)"
  ["lastChild"]=>
  string(22) "(object value omitted)"
  ["previousSibling"]=>
  NULL
  ["nextSibling"]=>
  string(22) "(object value omitted)"
  ["attributes"]=>
  string(22) "(object value omitted)"
  ["ownerDocument"]=>
  string(22) "(object value omitted)"
  ["namespaceURI"]=>
  NULL
  ["prefix"]=>
  string(0) ""
  ["localName"]=>
  string(3) "foo"
  ["baseURI"]=>
  string(29) "/Users/stloyd/Documents/flow/"
  ["textContent"]=>
  string(3) "bar"
}

Comment (norberttech) [reply src/core/etl/tests/Flow/ETL/Tests/Unit/Transformer/EntryExpressionEvalTransformerTest.php:97]: so how it was working before?

File: src/core/etl/src/Flow/ETL/Row/Entry/ArrayEntry.php [hunk 10]
```diff
@@ -76,6 +78,11 @@ public function name() : string
         return $this->name;
     }
 
+    public function phpType() : Type
+    {
+        return new ArrayType([] === $this->value);
```
norberttech: it would be more efficient to store phpType as class property created in constructor

src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -6,51 +6,39 @@
...
 
     /**
-     * JsonEntry constructor.
-     *
-     * @param array<mixed> $value
-     *
      * @throws InvalidArgumentException
      */
-    public function __construct(private readonly string $name, private readonly array $value)
+    public function __construct(private readonly string $name, string|array $value, private bool $object = false)
```
norberttech: what's the point of passing `$object` through constructor, when it's overwritten anyway?

src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -109,30 +94,39 @@ public function isEqual(Entry $entry) : bool
...
+    public function phpType() : Type
     {
-        $entry = new self($name, $this->value);
-        $entry->object = $this->object;
+        if ([] === $this->value) {
+            return ArrayType::empty();
+        }
 
-        return $entry;
+        return new ArrayType();
```
norberttech: I think this should be a StringType, because now on the one hand `phpType()` is array but `value()` return type is string, and `phpType()` should be aligned with return type of `value()`

src/core/etl/src/Flow/ETL/Row/Entry/ListEntry.php
```diff
@@ -24,26 +21,22 @@ final class ListEntry implements Entry, TypedCollection
...
         if ('' === $name) {
             throw InvalidArgumentException::because('Entry name cannot be empty');
         }
 
-        $this->type = new ListType($element);
-
-        if (!$this->type->isValid($value)) {
-            throw InvalidArgumentException::because('Expected list of ' . $element->toString() . ' got different types.');
+        if (!$type->isValid($value)) {
+            throw InvalidArgumentException::because('Expected ' . $type->toString() . ' got different types.');
```
norberttech: This does not say too much, I struggled with it before, exception was thrown but I had no idea what is wrong with the provided data structure, what I would do here is at use json encode on the value and add it to the exception message. But probably the best approach would be to use here TypeFactory, generate type and return toString() from it.

src/core/etl/tests/Flow/ETL/Tests/Unit/Row/Entry/StructureEntryTest.php
```diff
@@ -18,65 +22,140 @@ public static function is_equal_data_provider() : \Generator
...
+            new StructureEntry(
+                'name',
+                ['json' => ['5' => 5, '2' => 2, '1' => 1]],
+                new StructureType(new StructureElement('json', new ArrayType()))
+            ),
+            new StructureEntry(
+                'name',
+                ['json' => ['1' => 1, '2' => 2, '3' => 3]],
+                new StructureType(new StructureElement('json', new ArrayType()))
+            ),
```
norberttech, line 100: We going to need a few more tests here since we are no longer using Entries. 

- using nullable struct elements
- missing struct elements
- too many struct elements 
- invalid types struct elements

src/core/etl/src/Flow/ETL/Row/Schema/FlowMetadata.php
```diff
@@ -12,5 +12,7 @@ final class FlowMetadata
 
     public const METADATA_LIST_ENTRY_TYPE = 'flow_list_entry_type';
 
-    public const METADATA_STRUCTURE_DEFINITIONS = 'flow_structure_definitions';
+    public const METADATA_MAP_ENTRY_TYPE = 'flow_map_entry_type';
+
+    public const METADATA_STRUCTURE_ENTRY_TYPE = 'flow_structure_entry_type';
```
norberttech, line 19: ObjectEntry would need one as well since it already got Type.

src/core/etl/src/Flow/ETL/PHP/Type/Logical/List/ListElement.php
```diff
@@ -47,14 +47,14 @@ public static function map(MapType $type) : self
     /**
      * @param class-string $class
      */
-    public static function object(string $class) : self
+    public static function object(string $class, bool $optional = false) : self
```
norberttech: I think `nullable` is a better name, since this flag is not making list element optional, it allows it to be a null but it still needs to be present.

src/core/etl/src/Flow/ETL/Row/Entry.php
```diff
@@ -30,6 +31,8 @@ public function map(callable $mapper) : self;
 
     public function name() : string;
 
+    public function phpType() : Type;
```
norberttech: hmm or maybe just `type() : Type`? 🤔

src/core/etl/src/Flow/ETL/Row/Entry/EnumEntry.php
```diff
@@ -72,6 +74,11 @@ public function name() : string
         return $this->name;
     }
 
+    public function phpType() : Type
+    {
+        return ObjectType::fromObject($this->value);
```
norberttech: yeah, this requires EnumType 😉

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
```diff
@@ -30,7 +27,7 @@ public function getType(mixed $value) : Type
         }
 
         if (\is_object($value)) {
```
norberttech, line 30: hmm maybe `TypeDetector` would be a better name for this class, since it's not really creating anything new, it's detecting the type based on the provided value.

norberttech:
One more additional thing, MapEntry should be used across the test suite to make sure that it's working well with different transformers/loaders

src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -119,20 +110,25 @@ public function name() : string
...
+        return $this->type;
+    }
+
+    /**
+     * @throws \JsonException
+     */
     public function value() : string
     {
-        if (empty($this->value) && $this->object) {
+        if ([] === $this->value && $this->object) {
```
norberttech: `!\count($this->value)`

src/core/etl/src/Flow/ETL/Row/Entry/ListEntry.php
stloyd: Improved on list, map & structure, covered with some unit tests.

src/core/etl/src/Flow/ETL/PHP/Type/TypeFactory.php
stloyd: Renamed.

src/core/etl/tests/Flow/ETL/Tests/Unit/Transformer/EntryExpressionEvalTransformerTest.php
stloyd: Before there was a hack:
```php
            if ($class !== null) {
                if ($class === \DateTimeImmutable::class || $class === \DateTime::class) {
                    $class = \DateTimeInterface::class;
                }

                if ($class === \DOMElement::class) {
                    $class = \DOMNode::class;
                }

                return new Entry\ListEntry($entryName, ListElement::object($class), $value);
            }
```
Now it stores the class that was detected on an object.

src/core/etl/src/Flow/ETL/PHP/Type/Logical/StructureType.php
```diff
@@ -52,10 +52,12 @@ public function isEqual(Type $type) : bool
 
         foreach ($this->elements as $internalElement) {
             foreach ($type->elements() as $element) {
-                if (!$internalElement->isEqual($element->type())) {
-                    return false;
+                if ($internalElement->name() === $element->name() && $internalElement->isEqual($element->type())) {
+                    continue 2;
```
norberttech: This should just return true

norberttech (approved):

norberttech (approved):

PR #787, opened 2023-11-12
Removed DSL functions: `datetime_string()`, `json_string()`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Removed DSL functions: `datetime_string()`, `json_string()`</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Those functions were removed in favor of accepting string values in related DSL functions:
- `datetime_string()` => `datetime()`,
- `json_string()` => `json()` & `json_object()`

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Entry/DateTimeEntry.php
+++ b/src/core/etl/src/Flow/ETL/Row/Entry/DateTimeEntry.php
@@ -16,13 +16,25 @@ final class DateTimeEntry implements \Stringable, Entry
 {
     use EntryRef;
 
+    private readonly \DateTimeInterface $value;
+
     /**
      * @throws InvalidArgumentException
      */
-    public function __construct(private readonly string $name, private readonly \DateTimeInterface $value)
+    public function __construct(private readonly string $name, \DateTimeInterface|string $value)
     {
         if ($name === '') {
             throw InvalidArgumentException::because('Entry name cannot be empty');
+        }
+
+        if (\is_string($value)) {
+            try {
+                $this->value = new \DateTimeImmutable($value);
+            } catch (\Exception $e) {
+                throw new InvalidArgumentException("Invalid value given: '{$value}', reason: " . $e->getMessage(), previous: $e);
+            }
+        } else {
+            $this->value = $value;
         }
     }
 
@@ -44,7 +56,7 @@ final class DateTimeEntry implements \Stringable, Entry
 
     public function definition() : Definition
     {
-        return Definition::dateTime($this->name, false);
+        return Definition::dateTime($this->name);
     }
 
     public function is(string|Reference $name) : bool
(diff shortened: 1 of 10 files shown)


src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -11,28 +11,31 @@
...
 
     /**
-     * JsonEntry constructor.
-     *
-     * @param array<mixed> $value
-     *
      * @throws InvalidArgumentException
      */
-    public function __construct(private readonly string $name, private readonly array $value)
+    public function __construct(private readonly string $name, array|string $value)
```
norberttech, line 27: I think that `fromJsonString` can also be now safely removed as it has no purpose with constructor accepting union type

src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -132,7 +135,7 @@ public function toString() : string
 
     public function value() : string
     {
-        if (empty($this->value) && $this->object) {
+        if ([] === $this->value && $this->object) {
```
norberttech: `!\count($this->value)`

src/core/etl/src/Flow/ETL/Row/Entry/DateTimeEntry.php
```diff
@@ -16,14 +16,22 @@ final class DateTimeEntry implements \Stringable, Entry
...
      */
-    public function __construct(private readonly string $name, private readonly \DateTimeInterface $value)
+    public function __construct(private readonly string $name, \DateTimeInterface|string $value)
     {
         if ($name === '') {
             throw InvalidArgumentException::because('Entry name cannot be empty');
         }
+
+        if (\is_string($value)) {
+            $this->value = new \DateTimeImmutable($value);
```
norberttech: there should be a try-catch here in case someone will pass a value that is not a valid datetime string so we can throw consistent InvalidArgumentException

stloyd: Added also into json entry.

norberttech: Please add at least one unit tests for it, I would like to see how the exception message is going to look

norberttech (approved):

PR #788, opened 2023-11-12
Added BatchSizeOptimization

### LINKED ISSUE #738: Apply default batchSize based on Loader type
In order to improve DX, Flow should try to detect if the current BatchSize is suitable for a given Loader. 

```php

(new Flow())
    ->read(
        Dbal::from_limit_offset(
            $sourceDbConnection,
            'source_dataset_table',
            new OrderBy('id', Order::DESC)
        )
    )
    ->withEntry('id', ref('id')->cast('int'))
    ->withEntry('name', concat(ref('name'), lit(' '), ref('last name')))
    ->drop('last_name')
    ->write(Dbal::to_table_insert($dbConnection, 'flow_dataset_table'))
    ->run();
```

In this example, batchSize is equal to 1. This means that Flow will try to insert rows into the db, one by one. 
This can be easily changed by putting `batchSize(1_000)` just above `write`, but it also requires from developer some knowledge about how loaders work internally. 

What we can do, is use Optimizer in order to detect current batchSize when Loaders are added, and whenever we notice that batchSize wasn't set, we can automatically apply one. 
The exact numbers should be predefined, I think we can start from 1k for each of the following: 

- ElasticSearch
- Dbal
- Meilisearch 

For the file-based loaders, this is irrelevant, as most of them are writing rows one by one.

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>BatchSizeOptimization</li>
    <li>httpClient option to Meilisearch loader configuration</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Optimizer can be now configured through ConfigBuilder</li>
    <li>Execution Plan Processors can be now configured through ConfigBuilder</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes: #738

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "league/flysystem-aws-s3-v3": "^3.0",
         "league/flysystem-azure-blob-storage": "^3.0",
         "moneyphp/money": "^4",
-        "nyholm/psr7": "^1.4",
+        "nyholm/psr7": "^1.8",
         "php-http/curl-client": "^2.2",
         "php-http/mock-client": "^1.5",
         "ramsey/uuid": "^4.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "406d60d11a081d6deaa7757f9e697e1b",
+    "content-hash": "59f188fe625a8def83e99b55089edbbc",
     "packages": [
         {
             "name": "aeon-php/calendar",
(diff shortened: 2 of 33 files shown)


norberttech:
@stloyd I noticed that Meilisearch has two methods, bulk_index, and bulk_update, but the update does not seem to be used in any way. Do you think it should be removed?

stloyd:
@norberttech I think that it's leftover as the implementation started to be similar to ElasticSearch, but MeiliSearch has "better" handling of updates for multiple documents as it uses the same endpoint as for new ones.

I would leave it and open a new issue to validate if we want to have:
- add new **and update** existing with given IDs (current),
- or add new **and replace** existing with given IDs,

stloyd (approved):

PR #796, opened 2023-11-14
Remove Faker library from benchmark

### LINKED ISSUE #679: Improve bechmarks stability
We need to increase the stability & reproducibility of results provided by benchmarks.

To achieve this we need:
- [ ] replace faker usage with predefined "random" data setup,
- [x] reduce `revolutions` to lower value in all heavy benchmarks,
- [x] use `aggregate` report and have an eye at `rstdev` a value for giving info that benchmark is not stable enough (0-2% value are acceptable),

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Remove Faker library from benchmark</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes #679

### DIFF
--- a/src/core/etl/tests/Flow/ETL/Tests/Benchmark/EntryFactory/NativeEntryFactoryBench.php
+++ b/src/core/etl/tests/Flow/ETL/Tests/Benchmark/EntryFactory/NativeEntryFactoryBench.php
@@ -3,7 +3,6 @@
 namespace Flow\ETL\Tests\Benchmark\EntryFactory;
 
 use function Flow\ETL\DSL\array_to_rows;
-use Faker\Factory;
 use Flow\ETL\Row\Factory\NativeEntryFactory;
 use PhpBench\Attributes\Groups;
 use PhpBench\Attributes\ParamProviders;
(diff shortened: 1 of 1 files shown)


src/core/etl/tests/Flow/ETL/Tests/Benchmark/EntryFactory/NativeEntryFactoryBench.php
```diff
@@ -19,35 +18,30 @@ public function bench_entry_factory(array $params) : void
...
-                    'lng' => $faker->longitude,
+                    'lat' => ($i + 1) / 2,
+                    'lng' => ($i + 1) / 4,
                 ],
             ],
-            'notes' => \array_map(
-                static fn ($i) => $faker->sentence,
-                \range(1, 3)
-            ),
+            'notes' => \base64_encode(\random_bytes(\random_int(10, 200))),
```
norberttech: I think we should fix the dataset to constant values, randomizing values in benchmarks might give us random results, I would probably just put here static data since values does not matter in this case.

PR #800, opened 2023-11-19
Add integer 32 & 64 types

### LINKED ISSUE #771: Introduce `int32` and `int64` types in order to easier transform flow entries into parquet/avro types
> we might need to introduce here int32 and int64 in order to easier transform flow entries into parquet/avro types

_Originally posted by @norberttech in https://github.com/flow-php/flow/pull/770#discussion_r1386640334_

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add integer 32 & 64 types</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes #771

### DIFF
--- a/src/core/etl/tests/Flow/ETL/Tests/Unit/DataFrameTest.php
+++ b/src/core/etl/tests/Flow/ETL/Tests/Unit/DataFrameTest.php
@@ -1441,14 +1441,14 @@ ASCII,
         $this->assertStringContainsString(
             <<<'ASCII'
 schema
-|-- id: integer
+|-- id: integer64
 |-- country: string
-|-- age: integer
+|-- age: integer64
 schema
-|-- id: integer
+|-- id: integer64
 |-- country: string
-|-- age: integer
-|-- salary: ?integer
+|-- age: integer64
+|-- salary: ?integer64
 ASCII,
             $output
         );
--- a/src/core/etl/tests/Flow/ETL/Tests/Unit/DataFrameTest.php
+++ b/src/core/etl/tests/Flow/ETL/Tests/Unit/DataFrameTest.php
@@ -1636,7 +1636,7 @@ ASCII,
                 Row::create(Entry::integer('id', 2), Entry::string('name', 'bar'), Entry::boolean('active', false)),
             )
         )->validate(
-            new Schema(Schema\Definition::integer('id', $nullable = false)),
+            new Schema(Schema\Definition::integer('id', ScalarType::integer64())),
             new Schema\SelectiveValidator()
         )->fetch();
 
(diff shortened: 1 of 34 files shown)


norberttech:
int32/int64 should be also represented by schema definition

src/core/etl/tests/Flow/ETL/Tests/Unit/DataFrameTest.php
```diff
@@ -1714,9 +1714,9 @@ public function test_strict_validation_against_schema() : void
             )
         )->validate(
             new Schema(
-                Schema\Definition::integer('id', $nullable = false),
-                Schema\Definition::string('name', $nullable = true),
-                Schema\Definition::boolean('active', $nullable = false),
+                Schema\Definition::integer('id', ScalarType::integer32()),
```
norberttech, line 1717: Yeah, we need to decide here if we prefer to use `\Definition::integer()` or `\Definition::integer32()` - my problem with passing Type as a second parameter is that we need to validate it, which pretty much defeats the purpose of strongly typed arguments.
Alternatively, we can just drop all static factory methods (string/integer/boolean) and just use the constructor, however, that looks a bit less user-friendly. 

Additionally, I would put type into all Definition types, String, Boolean, Integer, we can also unify the metadata name to be something like: 

`METADATA_PHP_TYPE` which would always hold the instance of Type.

stloyd: I was not sure either how to decide about default that's why it's detecting inside what's max, but that's "magic", so I would go with BC break and force to choose either 32 or 64 on the user side but internally use 64 (similarly as it's now ie in parquet schema converter), with that I would also add check and throw exception when 64 is used when available is max 32.

So it would be: `Entry::integer64()`/`Entry::integer32()`, `Definition::integer64()` etc

Other than that I would save the helpers as DX is good there. Type for scalar is saved too.

norberttech: I was thinking about it over night, and I can't figure out any good use case for the end users to decide if they want to use int64 vs int63, yeah, one requires less space but with current storage prices I'm not sure how relevant that is. 
Originally I was planning to use that in order to know how to map Flow Int into Parquet Int and maybe we just keep it this way. Let me think more about it, I'm closing this PR for now, will get back to it

PR #824, opened 2023-11-22
Improve performance of `ArrayContentDetector`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Improve performance of `ArrayContentDetector`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

https://blackfire.io/profiles/compare/3bb42ece-7de3-489f-b640-44428c15380a/graph

## Summary by CodeRabbit

- **Refactor**
	- Improved the internal logic for content type detection in arrays to enhance performance and accuracy.
	- Removed an obsolete function related to array key validation, streamlining the class's public interface.

- **Bug Fixes**
	- Adjusted array handling methods to prevent potential misclassification of array types.

### DIFF
--- a/src/core/etl/src/Flow/ETL/PHP/Type/Native/ScalarType.php
+++ b/src/core/etl/src/Flow/ETL/PHP/Type/Native/ScalarType.php
@@ -96,11 +96,6 @@ final class ScalarType implements NativeType
         };
     }
 
-    public function isValidArrayKey() : bool
-    {
-        return \in_array($this->value, [self::INTEGER, self::STRING], true);
-    }
-
     public function nullable() : bool
     {
         return $this->nullable;
--- a/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
+++ b/src/core/etl/src/Flow/ETL/PHP/Type/ArrayContentDetector.php
@@ -11,49 +11,45 @@ use Flow\ETL\PHP\Type\Native\ScalarType;
 
 final class ArrayContentDetector
 {
+    private readonly ?Type $firstKeyType;
+
+    private readonly ?Type $firstValueType;
+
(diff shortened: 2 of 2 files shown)


src/core/etl/src/Flow/ETL/PHP/Type/Native/ScalarType.php
```diff
@@ -96,11 +96,6 @@ public function isValid(mixed $value) : bool
         };
     }
 
-    public function isValidArrayKey() : bool
```
stloyd, line 99: This check has no point as in PHP:
> The (array) key can either be an int or a string.

norberttech: yeah, but its too late at that point as php will throw exception, and this lets us know upfront

norberttech:
@coderabbitai review

src/core/etl/src/Flow/ETL/PHP/Type/Native/ScalarType.php
stloyd: At this point we are already with a PHP array, thus we will have int or string, and this check doesn't really do anything, the `ArrayContentDetect` works on the PHP array, so it cannot have type of keys other than int or string.

PR #829, opened 2023-11-23
Allow to replace ChartJS location

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Allow to replace ChartJS location</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
hey, so in general, I don't think that's the solution, this template is meant to be replaced with a custom one and not used on production. What would be probably more useful is a way to write that javascript content into a variable so it can be rendered without preparing any extra templates.

PR #837, opened 2023-11-25
Add `BatchLoader` interface for usage in `BatchSizeOptimization`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `BatchLoader` interface for usage in `BatchSizeOptimization`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Pointing explicitly with an interface which loaded can be used in `BatchSizeOptimization` allows us to prevent the circular dependency problem mentioned right now in that class.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Loader/BatchLoader.php
+++ b/src/core/etl/src/Flow/ETL/Loader/BatchLoader.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Loader;
+
+/**
+ * Loaders implementing BatchLoader interface points that they can use "batch size" optimization.
+ */
+interface BatchLoader
+{
+}
--- a/src/adapter/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/DbalLoader.php
+++ b/src/adapter/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/DbalLoader.php
@@ -27,7 +27,7 @@ use Flow\ETL\Rows;
  *  }
  * }>
  */
-final class DbalLoader implements Loader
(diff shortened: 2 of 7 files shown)


src/core/etl/src/Flow/ETL/Loader/BatchLoader.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types=1);
+
+namespace Flow\ETL\Loader;
+
+/**
+ * Loaders implementing BatchLoader interface points that they can use "batch size" optimization.
+ */
+interface BatchLoader
+{
+}
```
norberttech, line 12: I'm not a huge fan of empty interfaces without any behavior, it just feels wrong

stloyd: Yeah, true to that, but other ways:
- extend the normal `Loader` interface, but this will still be empty...
- add some method returning a boolean, but sounds smelly too,
- add a new method in the `Loader` interface returning boolean, ie. `supportsBatchLoading() : bool`

norberttech: Another problem is that we also have batching pipelines that are being excluded here and that are also stored as a list of classes. For them, we would need to create another empty interface. 
I think I'm better keeping those classes as they are now

Issue #567, opened 2023-10-11
Schema inferring


norberttech:
When working on DataFrames without provided schema, Flow does its best to detect the entry value type to create the proper Entry type in NativeEntryFactory. 

This behavior is very developer-friendly but introduces a significant impact on Extractors using `array_to_rows()`

This could be improved by introducing schema inferring similar to the one available in Apache Spark. 

We could extract the first X rows (this should be configurable), iterate over them, and detect the schema. 
Then, revert the extractor and read all rows passing inferred schema to NativeEntryFactory, which would no longer need to guess entry types.

norberttech:
I'm closing this one since DataFrame::autoCast brings Schema Inferring to Flow. At that point, I'm not sure if there is any value in wrapping this into some independent mechanism. That way it's much more flexible.

PR #839, opened 2023-11-26
Add `Schema` into the `FlowContext`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `Schema` into the `FlowContext`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/adapter/etl-adapter-avro/src/Flow/ETL/Adapter/Avro/FlixTech/AvroExtractor.php
+++ b/src/adapter/etl-adapter-avro/src/Flow/ETL/Adapter/Avro/FlixTech/AvroExtractor.php
@@ -49,7 +49,7 @@ final class AvroExtractor implements Extractor, Extractor\FileExtractor, Limitab
                     $row['_input_file_uri'] = $filePath->uri();
                 }
 
-                $signal = yield array_to_rows($row, $context->entryFactory(), $partitions);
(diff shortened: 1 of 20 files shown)


norberttech:
hmm, I'm not fully sure if I understand the intention here, why would we like to expose schema to context?

stloyd:
This was my idea for the introduction of `SchemaDetectionPipeline` (really draft idea), which would first iterate over X rows, setup schema, and then pass it into the context & re-iterate, something like: https://gist.github.com/stloyd/8393a3ec270dd69dddd228bbb3bfc159#file-schemadetectionpipeline-php-L79

But as said it's really a draft idea on touching the iceberg of #567 😄

norberttech:
oh I see, so here is the problem with this approach. Schema is too dynamic to be a part of execution context, context is for things that are not changing or that are transformation independent. You would need to update the schema in the context after every single step, otherwise it might change before it even reaches the point where it should be used.

PR #848, opened 2023-11-28
Use platform for column escaping in bulk insert

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Use platform for column escaping in bulk insert</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Ensure that column names in generated queries are escaped to match the default behavior done in Doctrine.

### DIFF
--- a/src/lib/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/Columns.php
+++ b/src/lib/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/Columns.php
@@ -34,11 +34,6 @@ final class Columns
         return $this->columns;
     }
 
-    public function concat(string $separator) : string
-    {
-        return \implode($separator, $this->columns);
-    }
-
     /**
      * @param string ...$columnNames
      *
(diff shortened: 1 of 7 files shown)


src/lib/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/Columns.php
```diff
@@ -34,11 +34,6 @@ public function all() : array
         return $this->columns;
     }
 
-    public function concat(string $separator) : string
```
norberttech, line 37: why?

norberttech (approved):

src/lib/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/Columns.php
stloyd: It's not used anymore, we can leave it ofc, but it's only usage was to concat with `,` which now is done with quoting on the level of dialect classes, not in `Columns`.

PR #849, opened 2023-11-28
Rework Doctrine Bulk tests to not use deprecated code

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Rework Doctrine Bulk tests to not use deprecated code</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/lib/doctrine-dbal-bulk/tests/Flow/Doctrine/Bulk/Tests/SqliteIntegrationTestCase.php
```diff
@@ -4,21 +4,37 @@
...
+            )
+        );
     }
 
     protected function tearDown() : void
     {
         $this->databaseContext->dropAllTables();
     }
+
+    public function numberOfExecutedInsertQueries() : int
```
norberttech: `executedQueriesCount() : int`

src/lib/doctrine-dbal-bulk/tests/Flow/Doctrine/Bulk/Tests/MysqlIntegrationTestCase.php
```diff
@@ -4,22 +4,38 @@
...
     protected function setUp() : void
     {
-        $this->databaseContext = new DatabaseContext(DriverManager::getConnection(['url' => \getenv('MYSQL_DATABASE_URL')]));
+        $this->logger = new ProxyLogger();
+        $this->databaseContext = new DatabaseContext(
+            DriverManager::getConnection(
+                ['url' => \getenv('MYSQL_DATABASE_URL')],
+                (new Configuration())->setMiddlewares([new Middleware($this->logger)])
+            )
+        );
```
norberttech, line 21: <s>I think I would extract logger setting into IntegrationTestCase and make all dialects to extend from it. </s> wait, I just noticed that it was removed from the Context, why?

stloyd: The reason is the `DriverManager::getConnection` here, it's setting up the configuration (with default setup), which later (in context) cannot be replaced and always uses default one, thus I needed to move this level up.

PR #869, opened 2023-12-07
RDSL lib init commit

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>RDSL library</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

RDSL is a simple library that allows to define a chain of execution of DSL functions. 
It comes with access control lists, to exclude/include only specific functions from given namespaces to be a part of the DSL.
It also allows to define which functions are available as an entry points, also through ACL. 

Usage:

```php
<?php

$executables = $builder->parse(
    [
        [
            'function' => 'int',
            'args' => [0],
            'call' => [
                'method' => 'add',
                'args' => [
                    [
                        'function' => 'lit',
                        'args' => [5],
                    ],
                ],
            ],
        ],
    ]
);

$results = (new Executor())->execute($executables);
```

The goal is to define data_frame/df functions as an entry point from which we can call other functions. 
ETL will parse pipelines from Yaml/Json/XML, turn them into associative arrays and build DataFrame. 

It will all be possible through a static factory `DataFrame::parse(array $definition) : DataFrame`. 
Then specific parses would just use that static factory, for example `Flow\ETL\Parser\JsonParser::parse(string $json)`

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -46,6 +46,7 @@ jobs:
           tools: composer:v2
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
+          extensions: :psr
 
       - name: "Create cache directories"
         run: |
@@ -170,7 +171,7 @@ jobs:
           tools: composer:v2
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
-          extensions: none, bcmath, curl, dom, fileinfo, filter, hash, json, mbstring, pcre, simplexml, xmlreader, zlib
+          extensions: :psr
 
       - name: "List PHP Extensions"
         run: php -m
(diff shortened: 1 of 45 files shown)


.github/workflows/test-suite.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           tools: composer:v2
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
+          extensions: :psr
```
norberttech, line 49: I'm honestly not sure why suddently psr extension become installed but it was breaking the testsuite, disabled according to [docs](https://github.com/marketplace/actions/setup-php-action#heavy_plus_sign-php-extension-support)

Issue #793, opened 2023-11-14
Remove AsyncPipeline


norberttech:
Our current implementation of asynchronous processing is built on top of React/AMPhp. Flow is simply spawning new processes and it's sending the pipeline into it, later those processes are receiving rows over TCP, transforming them and sending back also through TCP to the main process. 

This approach, even that is pretty cool is not brining enough value in order to justify the maintenance around it. 
The main problem comes from the fact that processing is not the biggest bottleneck, our main problem comes from extraction, which is not included in our current async processing model. 

Another problem is that async processing is currently impossible through phar/docker. 

For now we are going to remove the AsyncPipeline and react/amphp adapters, which should reduce the complicity of our codebase allowing us to easier refactor Pipelines (we are still missing Executor). One that is done, we should look again into async processing, however this time we need to start from async extraction.

PR #891, opened 2023-12-30
Adjust serializer interface

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Remove compression serializer</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

The code was related to async code removed in https://github.com/flow-php/flow/pull/794 related to #793


norberttech (changes requested):
Serializable interface with all __serialize() and __unserialize methods should also be removed (maybe Rector can help to autome it). 

But for the Cache I think I would still keep a thin abstraction for Serializer with native implementation that uses \serialize method. 

Something like: 

```php
Serializer::serialize(object $object) : string
Serializer::unserialize(string $serialized, string $class) : object
```

src/core/etl/src/Flow/Serializer/NativePHPSerializer.php
```diff
@@ -17,14 +17,18 @@ public function serialize(Serializable $serializable) : string
...
-    public function unserialize(string $serialized) : Serializable
+    public function unserialize(string $serialized, string $class) : Serializable
     {
         $value = \unserialize($serialized, ['allowed_classes' => true]);
 
         if (!$value instanceof Serializable) {
             throw new RuntimeException('NativePHPSerializer::unserialize must return instance of Serializable');
         }
 
+        if (!\is_a($value, $class)) {
```
norberttech, line 28: I think you might want to pass class name to allowed_class option and then simplify that condition.

norberttech (approved):

PR #903, opened 2024-01-02
Add XML writer

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add XML writer</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/XMLLoader.php
```diff
@@ -0,0 +1,112 @@
...
+        $xml = Writer::inMemory()
+            ->apply(
+                static function (\XMLWriter $xmlWriter) use ($rows) : void {
+                    $data = [];
+
+                    foreach ($rows->toArray() as $row) {
+                        $children = [];
+
+                        foreach ($row as $name => $value) {
+                            $children[] = element($name, value((string) $value));
```
stloyd: `value()` method from Vee requires getting a string, but XML allows integers and other values.

src/core/etl/src/Flow/ETL/Row/Entry/XMLEntry.php
```diff
@@ -119,7 +118,7 @@ public function rename(string $name) : Entry
     public function toString() : string
     {
         /** @phpstan-ignore-next-line */
-        return $this->value->saveXML();
+        return $this->value->saveXML($this->value->documentElement);
```
stloyd, line 121: This removes additional `<?xml version="1.0" ?>` being added to nodes.

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/XMLLoader.php
```diff
@@ -0,0 +1,112 @@
...
+        if (!\count($rows)) {
+            return;
+        }
+
+        $xml = Writer::inMemory()
+            ->apply(
+                static function (\XMLWriter $xmlWriter) use ($rows) : void {
+                    $data = [];
+
+                    foreach ($rows->toArray() as $row) {
```
norberttech: I think I would extract this entire logic to RowsNormalizer, just like with parquet and follow similar logic with matching the entry type and using ->toString() since ->value() might not always be accepted.

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/XMLLoader.php
norberttech: wait, how does xml allow for integers? `<node>1</node>` and `<node>test</node>` are always string, even that first one is numeric, it's the schema that defines allowed values, not the file itself, no?

src/adapter/etl-adapter-xml/composer.json
```diff
@@ -14,7 +14,9 @@
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
         "ext-dom": "*",
         "ext-xmlreader": "*",
-        "flow-php/etl": "^0.5.0 || 1.x-dev"
+        "ext-xmlwriter": "*",
+        "flow-php/etl": "^0.5.0 || 1.x-dev",
+        "veewee/xml": "^2.13.0"
```
norberttech: we should aim to support lowest version of the lib, so if none of our features are using things added after 2.0.0 we should rather set this dependency to be `^2.0`, unless we are using something added/fixed in `2.13.0.` then `^2.13.0`

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/XMLLoader.php
```diff
@@ -0,0 +1,112 @@
...
+        $streams = $context->streams();
+
+        if (!$streams->isOpen($this->path, $partitions)) {
+            $stream = $streams->open($this->path, 'xml', $context->appendSafe(), $partitions);
+
+            if (!\array_key_exists($stream->path()->uri(), $this->writes)) {
+                $this->writes[$stream->path()->uri()] = 0;
+            }
+
+            \fwrite($stream->resource(), '<rows>');
```
norberttech: this does not look right, I thought that the lib provides some way to generate entire file without manually adding HTML? 

Also don't we need to first add `<?xml version="1.0" encoding="UTF-8"?>`?

src/core/etl/src/Flow/ETL/Row/Entry/XMLEntry.php
norberttech: Why do we want to remove it?

stloyd: As you can see in tests, by default `\DOMDocument` adds `<?xml` tags to everything, even if you extract a particular node of the XML which looks invalid IMO.

src/adapter/etl-adapter-xml/composer.json
stloyd: Removed usage of the library in favor of direct playing with `DOMDocument` & `XMLWriter`.

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/Loader/DomDocumentLoader.php
```diff
@@ -17,9 +17,9 @@ final class DomDocumentLoader implements Closure, Loader, Loader\FileLoader
 {
     public function __construct(
         private readonly Path $path,
-        private readonly RowsNormalizer $normalizer = new RowsNormalizer(),
         private readonly string $collectionName = 'rows',
         private readonly string $collectionElementName = 'row',
+        private readonly RowsNormalizer $normalizer = new RowsNormalizer(),
```
norberttech, line 22: honestly I dont think we want to expose that, I don't see any reason why anyone should touch normalizers.

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/Loader/DomDocumentLoader.php
```diff
@@ -0,0 +1,83 @@
...
+use Flow\ETL\FlowContext;
+use Flow\ETL\Loader;
+use Flow\ETL\Loader\Closure;
+use Flow\ETL\Rows;
+
+final class DomDocumentLoader implements Closure, Loader, Loader\FileLoader
+{
+    public function __construct(
+        private readonly Path $path,
+        private readonly string $collectionName = 'rows',
```
norberttech, line 20: `private readonly string $rootElement = 'rows',`

src/adapter/etl-adapter-xml/src/Flow/ETL/Adapter/XML/Loader/DomDocumentLoader.php
```diff
@@ -0,0 +1,83 @@
...
+        private readonly string $collectionElementName = 'row',
```
norberttech, line 21: `private readonly string $rowElement = 'row'`

norberttech:
I see one more issue with those Loaders, there is no way to write anything as an attribute 🤔 We should try to came up with at least the idea how such a feature could be implemented before we merge this in order to avoid future BC breaks. 

 I can totally see how someone would like to say that for example on `<user> (row name)` he would like entry `id` to be an attribute `<user id="xxxx"><email>something@email.com</email></user>`

norberttech:
> I see one more issue with those Loaders, there is no way to write anything as an attribute 🤔 We should try to came up with at least the idea how such a feature could be implemented before we merge this in order to avoid future BC breaks.
> 
> I can totally see how someone would like to say that for example on `<user> (row name)` he would like entry `id` to be an attribute `<user id="xxxx"><email>something@email.com</email></user>`

@stloyd any thoughts?

stloyd:
> I see one more issue with those Loaders, there is no way to write anything as an attribute 🤔 We should try to came up with at least the idea how such a feature could be implemented before we merge this in order to avoid future BC breaks. 
> 
> 
> 
>  I can totally see how someone would like to say that for example on `<user> (row name)` he would like entry `id` to be an attribute `<user id="xxxx"><email>something@email.com</email></user>` 

While setting XML attributes may makes sense, I would not go further in this basic implementation with it cause we don't have real use cases (for sure I don't have).

For me that implementation is good start, later on we add some helpers/transformers that would allow support for attributes creation.

The only thing I would implement in this code would be change in rows normalisation for XMLEntry to not escape it but write directly as raw.

norberttech:
> > I see one more issue with those Loaders, there is no way to write anything as an attribute 🤔 We should try to came up with at least the idea how such a feature could be implemented before we merge this in order to avoid future BC breaks.
> > I can totally see how someone would like to say that for example on `<user> (row name)` he would like entry `id` to be an attribute `<user id="xxxx"><email>something@email.com</email></user>`
> 
> While setting XML attributes may makes sense, I would not go further in this basic implementation with it cause we don't have real use cases (for sure I don't have).
> 
> For me that implementation is good start, later on we add some helpers/transformers that would allow support for attributes creation.
> 
> The only thing I would implement in this code would be change in rows normalisation for XMLEntry to not escape it but write directly as raw.

Sorry, that wasn't my question. I was asking not about if it should be implemented but how. 

Here are the blockers without which I would not like to merge this XML Loader. 

- support for saving specific entries as attributes 
- support for all logical types (recursive handling of maps/structs, etc) 

In the current state, we can merge it into a feature branch if you prefer working in smaller iterations.
I was playing a bit with this adapter, and without those two things, it's pretty much useless for any real-world use cases. It's extremely frustrating to flatten every single data structure. Not using attributes also impacts the total file size. 

Let me know if you would like to continue working on this on a feature branch or close it for now and open it again later if you manage to figure out those two things.

norberttech:
I'm closing this from now due to lack of activity, if you would like to pick this up again later, feel free to open a new PR or use this one

PR #910, opened 2024-01-08
add csv adapter unit test for fixture not yet in use

mzarnecki:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>add csv adapter unit test for fixture not yet in use</li> 
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-csv/tests/Flow/ETL/Adapter/CSV/Tests/Integration/CSVExtractorTest.php
```diff
@@ -263,6 +263,14 @@ public function test_limit() : void
         );
     }
 
+    public function test_load_json_file_as_csv_gives_empty_rows_list() : void
+    {
+        $extractor = from_csv(Path::realpath(__DIR__ . '/../Fixtures/not_csv.csv'));
+        $generator = $extractor->extract(new FlowContext(Config::default()));
+
+        $this->assertEmpty(\iterator_to_array($generator));
```
mzarnecki: is it intentional that empty set is returned in such case or should it be handled by throwing dedicated exception?

norberttech: Great question, let me look into this and get back to you a bit later

norberttech: Ok, I know what's going on, the reason why this is not failing is in the implementations of `Flow\ETL\Filesystem` interface. 

- [FlysystemFS::scan()](https://github.com/flow-php/flow/blob/1.x/src/adapter/etl-adapter-filesystem/src/Flow/ETL/Adapter/Filesystem/FlysystemFS.php#L133) 
- [LocalFilesystem::scan()](https://github.com/flow-php/flow/blob/1.x/src/core/etl/src/Flow/ETL/Filesystem/LocalFilesystem.php#L113) 

Those two functions should have the following condition: 

```php
        if (!$path->isPattern() && !$this->fileExists($path)) {
            throw new RuntimeException(\sprintf('Path "%s" does not exists', $path->uri()));
        }
```

This means that if the path is not a pattern and the path does not exist, it should throw an exception.    

This will break tests like this one: 

```php
$path = \sys_get_temp_dir() . '/partitioned';
        $this->cleanDirectory($path);

        df()
            ->read(from_rows($rows = new Rows(
                $this->createRow(1, new \DateTimeImmutable('2020-01-01 00:01:00')),
                $this->createRow(1, new \DateTimeImmutable('2020-01-01 00:02:00')),
                $this->createRow(1, new \DateTimeImmutable('2020-01-02 00:01:00')),
                $this->createRow(1, new \DateTimeImmutable('2020-01-02 00:02:00')),
                $this->createRow(1, new \DateTimeImmutable('2020-01-03 00:01:00')),
            )))
            ->withEntry('date', ref('datetime')->toDate()->dateFormat())
            ->partitionBy(ref('date'))
            ->write(to_parquet($path))
            ->run();

        $this->assertEquals(
            $rows,
            (new Flow())
                ->read(from_parquet($path))
                ->drop('date')
                ->sortBy(ref('datetime')->asc())
                ->fetch()
        );
```

but that's desired outcome, because those tests are broken, this is how it should look: 

```diff
- ->read(from_parquet($path))
+ ->read(from_parquet($path . '/*.parquet'))
```

which is valid pattern and not just a folder. 

It's all because originally I was assuming that `Filesystem::scan(path('/path/to/folder')` should scan the entire directory, but since scan supports glob, it's inconsistent and should be fixed just like explained above. 

Great catch!

mzarnecki: you're right
thanks for very detailed explanation!

src/adapter/etl-adapter-filesystem/src/Flow/ETL/Adapter/Filesystem/FlysystemFS.php
```diff
@@ -132,6 +132,10 @@ public function rm(Path $path) : void
      */
     public function scan(Path $path, PartitionFilter $partitionFilter = new NoopFilter()) : \Generator
     {
+        if (!$path->isPattern() && !$this->fileExists($path) && !$this->directoryExists($path)) {
```
norberttech: I'm not sure if we should keep that `!$this->directoryExists($path)` since it's creating that weird behavior that I mentioned before. 

`/path/to/dir` will scan the entire dir, just like `/path/to/dir/**/*` 

And I think that since we support glob, auto scanning entire directory by `scan` method might not be necessary.

norberttech: hey @rzarno 
Are there any issues with removing that extra condition?

mzarnecki: hey Norbert,
yes, I originally made it without directory related condition as you suggested  and some additional unit tests broke  
so I need to take a deeper look, but didn't had time yet

norberttech: sounds good, I just wanted to check if your aren't stuck somewhere with this. Those tests should be pretty straightforward to fix as you just need to adjust the paths from `/path/to/folder` into something like this `/path/to/folder/**/*.json`.
But if there would be any edge cases that I'm missing now, please let me know

mzarnecki: yes, indeed, wildcards were enough to solve issues with units
updated PR, thanks for advice

norberttech:
Thanks, @rzarno, great catch!

PR #919, opened 2024-01-18
Tiny performance improvement on Path::partitions

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>optimized Path::partitions method by using mnemonic</li>
    <li>updated dependencies</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


bin/flow.php
```diff
@@ -16,7 +16,7 @@
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
-if ('' !== \Phar::running(false)) {
+if ('' !== Phar::running(false)) {
```
norberttech, line 19: I mean, I have no idea why CS Fixer is doing that but at this point I don't care, I spent way too much time trying to understand what/why cs fixer is doing something, it become extremely unstable recently, I might need to lock it to some specific version.

PR #923, opened 2024-01-18
Added DataFrame::autoCast()

### LINKED ISSUE #921: Auto Cast / Auto Detect Types
Right now, when loading data from formats like CSV or JSON everything is converted into StringEntry, 
it's because those formats does not have any kind of schema. 
It would be very beneficial to create a mechanism that when used would try to guess best possible type based on data value and cast entry into this. 

For example: 

- `"true"` -> boolean
- `"1"` -> int 
- `"2023-03-04"` -> DateTime 

I see it as a custom transformer, something that can be used like this: 

```
df()
   ->read(from_csv('...'))
   ->autoCast() 
   ->write(to_parquet('...'))
   ->run(); 
```

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>DataFrame::autoCast()</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Extracted isJson,isXML,isUuif from NativeEntryFactory to StringTypeChecker</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

Closes: #921 

<h2>Description</h2>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Transformer/AutoCastTransformer.php
+++ b/src/core/etl/src/Flow/ETL/Transformer/AutoCastTransformer.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Transformer;
+
+use function Flow\ETL\DSL\bool_entry;
+use function Flow\ETL\DSL\datetime_entry;
+use function Flow\ETL\DSL\float_entry;
+use function Flow\ETL\DSL\int_entry;
+use function Flow\ETL\DSL\json_entry;
+use function Flow\ETL\DSL\null_entry;
+use function Flow\ETL\DSL\uuid_entry;
+use Flow\ETL\FlowContext;
+use Flow\ETL\Row;
+use Flow\ETL\Row\Entry;
+use Flow\ETL\Row\Entry\StringEntry;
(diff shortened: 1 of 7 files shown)


src/core/etl/src/Flow/ETL/Transformer/AutoCastTransformer.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        return $entry;
+    }
+
+    public function transform(Rows $rows, FlowContext $context) : Rows
+    {
+        return $rows->map(function (Row $row) {
+            return $row->map(function (Entry $entry) {
+                return $this->autoCast($entry);
```
norberttech, line 66: We might want to add later some additional caching layer here, that would keep a type and entry after first casting to avoid going through the same detection process over and over.
Of course it's still possible that not all rows would get the same type, but in that case we can catch exception and fallback to autoCast function.

PR #949, opened 2024-02-02
Synchronize web `composer.json` with others

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Synchronize web `composer.json` with others</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/monorepo-builder.php
+++ b/monorepo-builder.php
@@ -11,6 +11,7 @@ return static function (MBConfig $config): void {
         __DIR__ . '/src/core',
         __DIR__ . '/src/adapter',
         __DIR__ . '/src/lib',
+        __DIR__ . '/src/web',
     ]);
 
     $config->dataToAppend([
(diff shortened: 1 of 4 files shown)


monorepo-builder.php
```diff
@@ -11,6 +11,7 @@
         __DIR__ . '/src/core',
         __DIR__ . '/src/adapter',
         __DIR__ . '/src/lib',
+        __DIR__ . '/src/web',
```
norberttech, line 14: hmm the way I was thinking about it was that we are going to build static content and only that static content would be published to standalone repo that would be gh static page. 

I'm also not sure if we even want to keep the same composer.json for landing page, it's not that this landing page should use the same dependencies and the code, no?

PR #1063, opened 2024-04-29
fixes #1062 : removed empty aws client credentials and set default region

### LINKED ISSUE #1062: AWS S3 client credentials should be left optional
When deploying PHP code on an AWS infrastructure where grants are set upon IAM roles, credentials are not required. Though, Flow\ETL\Adapter\Filesystem\AwsS3Stream define them as empty by default.
Same goes with region.

jguittard:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>removed empty aws client credentials and set default region</li> 
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/adapter/etl-adapter-filesystem/src/Flow/ETL/Adapter/Filesystem/AwsS3Stream.php
+++ b/src/adapter/etl-adapter-filesystem/src/Flow/ETL/Adapter/Filesystem/AwsS3Stream.php
@@ -42,11 +42,7 @@ final class AwsS3Stream extends FlysystemWrapper
              */
             $clientOptions = \array_merge(
                 [
-                    'credentials' => [
-                        'key' => '',
-                        'secret' => '',
-                    ],
-                    'region' => '',
+                    'region' => 'us-east-1',
                     'version' => 'latest',
                 ],
                 $contextOptions[self::PROTOCOL]['client'] ?? []


src/adapter/etl-adapter-filesystem/src/Flow/ETL/Adapter/Filesystem/AwsS3Stream.php
```diff
@@ -42,11 +42,7 @@ protected function filesystem() : Filesystem
              */
             $clientOptions = \array_merge(
                 [
-                    'credentials' => [
-                        'key' => '',
-                        'secret' => '',
-                    ],
-                    'region' => '',
+                    'region' => 'us-east-1',
```
norberttech, line 45: I'm honestly not sure about default region, I think I would rather prefer to throw an exception when the region is not set. Is there anything that makes us-east-1 special enough to make it default?

jguittard: It's just because it's the default region AWS applies.
IMHO I'd rather avoid merging any default configuration :)

norberttech: > IMHO I'd rather avoid merging any default configuration :)

Same, can't say out of my head why it was implemented this way, I need to take another look later in the codebase but as far as I remember there was a reason. I will look into this later and get back to you

jguittard: Simple : S3 client needs an array with at least `region` and `version` set. If `$contextOptions` does not provide it, it will throw an exception within aws-sdk. Maybe check those 2 keys right before instantiating `Filesystem` and throw a domain exception instead?

norberttech: I was playing with it a bit and I think we can skip any defaults, when region or credentials are missing, S3 client will throw an exception which I think is good since it's S3 responsibility, not flow. 
Could you remove those merges from both, S3 and Azure clients and in the meantime I'm gonna prepare some examples for the docs that would explain how to use them both?

jguittard: You got it. I'll take care of it tomorrow

norberttech: sounds good, examples should be available any moment: https://github.com/flow-php/flow/pull/1065

norberttech: @jguittard would you mind if I would take it over from here and adjust your PR so it can be merged?

PR #1068, opened 2024-05-06
Replace `XMLNodeEntry` with new `XMLElementEntry`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add new XMLElementEntry</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Remove XMLNodeEntry</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">

  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Using simply just `DOMNode` generates problems as we're unable to work with XML attributes, which are available on a class that extends `DOMNode` = `DOMElement`.


src/core/etl/tests/Flow/ETL/Tests/Unit/Row/Entry/XMLElementEntryTest.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Flow\ETL\Tests\Unit\Row\Entry;
+
+use Flow\ETL\Row\Entry\XMLElementEntry;
+use PHPUnit\Framework\TestCase;
+
+final class XMLElementEntryTest extends TestCase
+{
+    public function test_serialization() : void
```
norberttech, line 35: We need more tests here for the initialization of this DOMElement from object and string, additionally isEqual and toString since they also holds some additional logic.

norberttech:
I would simply remove DOMNodeEntry and replace it with DOMElementEntry, I fucked up a bit at the beginning not reading properly definition of DOMNode which can be anything, including the attribute, DOMElement is what I was actually thinking about but since it wasn't usable I don't think it's a huge BC Break. 

About attributes, I think it should be handled out of the scope of this PR since it's more XMLWriter responsibility rather than DOMElementEntry to convert entries into elements/attributes. 

`\Flow\ETL\Function\XPath` you might need to look into this one as well since it returns DOMNode which is later converted by the native entry factory into DOMNodeEntry but I think it should be pretty straightforward to cast it into DOMElement. Afaik it's the only scalar function that operates on XML.

src/core/etl/src/Flow/ETL/Function/DOMElementAttribute.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Function;
+
+use Flow\ETL\Row;
+
+final class DOMElementAttribute extends ScalarFunctionChain
```
norberttech, line 9: another good scalar function would be DOMElementHasAttribute which could be user together with `when()` expression. like `when(ref('node')->domElementHasAttribute("id"))->then(...)`, but it's out of the scope of that PR.

PR #1073, opened 2024-05-10
Adjust Inflection to run only on changed files

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Adjust Inflection to run only on changed files</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Using new feature from https://github.com/infection/infection/pull/1919 to not run Inflection against whole codebase, but only on the changed files.

### DIFF
--- a/.github/workflows/test-mutations.yml
+++ b/.github/workflows/test-mutations.yml
@@ -77,4 +77,6 @@ jobs:
         run: "composer install --no-interaction --no-progress --no-suggest"
 
       - name: "Mutation Tests"
-        run: "composer test:mutation -- --logger-github=false"
+        run: |
+          git fetch origin $GITHUB_BASE_REF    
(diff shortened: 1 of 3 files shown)


norberttech:
don't we need to add also `#CoversClass` in order to make it work?

stloyd:
Unfortunately yes... so we cannot bring this improvement...

norberttech:
could you leave the other adjustments, like using max cores? I was going to add that but there is always something more important on my plate

norberttech:
@stloyd maybe there is an option to automate that https://github.com/rectorphp/rector-phpunit/pull/319#issue-2204576592

PR #1105, opened 2024-07-03
Add support for ZSTD compression

### LINKED ISSUE #782: Add support for ZSTD compression
Extension: https://github.com/kjdev/php-ext-zstd

flavioheleno:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>zstd compression functions</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

This PR adds support for ZSTD compression in Parquet Codec.

Closes #782.

### DIFF
--- a/src/lib/parquet/src/Flow/Parquet/ParquetFile/Codec.php
+++ b/src/lib/parquet/src/Flow/Parquet/ParquetFile/Codec.php
@@ -26,6 +26,7 @@ final class Codec
             Compressions::SNAPPY => \snappy_compress($data),
             /** @phpstan-ignore-next-line */
             Compressions::GZIP => \gzencode($data, $this->options->get(Option::GZIP_COMPRESSION_LEVEL)),
+            Compressions::ZSTD => \zstd_compress($data, $this->options->getInt(Option::ZSTD_COMPRESSION_LEVEL)),
             default => throw new RuntimeException('Compression ' . $compression->name . ' is not supported yet')
         };
 
@@ -43,6 +44,7 @@ final class Codec
             Compressions::UNCOMPRESSED => $data,
             Compressions::SNAPPY => \snappy_uncompress($data),
             Compressions::GZIP => \gzdecode($data),
+            Compressions::ZSTD => \zstd_uncompress($data),
             default => throw new RuntimeException('Compression ' . $compression->name . ' is not supported yet')
         };
 
--- a/src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/CompressionTest.php
+++ b/src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/CompressionTest.php
@@ -8,6 +8,7 @@ use Faker\Factory;
 use Flow\Parquet\ParquetFile\Schema\{FlatColumn, ListElement, NestedColumn};
 use Flow\Parquet\ParquetFile\{Compressions, Schema};
 use Flow\Parquet\{Consts, Reader, Writer};
+use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\TestCase;
 
 final class CompressionTest extends TestCase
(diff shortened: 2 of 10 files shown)


src/lib/zstd/polyfill.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
norberttech: I think I would move that file to parquet lib, since in order to make a valid polyfill we would need to provide a pure PHP implementation as I did with snappy. Until then I would keep this as src/lib/parquet/stubs.php or something similar, just don't forget to add it to parquet composer.json autoload and main project composer.json autoload

src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/CompressionTest.php
```diff
@@ -152,4 +152,55 @@ public function test_writing_and_reading_file_with_uncompressed_compression() :
         self::assertFileExists($path);
         \unlink($path);
     }
+
+    public function test_writing_and_reading_file_with_zstd_compression() : void
+    {
+        if (!\extension_loaded('zstd')) {
```
norberttech, line 160: since zstd is not available by default what I would do is I would put in phpunit.xml.dist a [section](https://phpunit.de/manual/6.5/en/organizing-tests.html#organizing-tests.xml-configuration) for tests that depends on zstd. Next I would try to create another github workflow that would compile and install that extension just to run those tests - but if you don't feel like working on it, you can keep it as is and I will add the zstd testsutie later

flavioheleno: should the section be specific, like:
```xml
<testsuite name="zstd">
  <file>src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/CompressionTest.php</file>
  <file>src/lib/parquet/tests/Flow/Parquet/Tests/Unit/ParquetFile/CodecTest.php</file>
</testsuite>
```

or more broad like the other sections, like:

```xml
<testsuite name="zstd">
  <directory>src/lib/**/**/**/**/Tests/Integration</directory>
  <directory>src/lib/**/**/**/**/Tests/Unit</directory>
</testsuite>
```

norberttech: I would call the testsuite `zstd-extension` and pull those tests into standalone files like: 

- `src/lib/parquet/tests/Flow/Parquet/Tests/Integration/IO/Compression/ZSTDTest.php`
- `src/lib/parquet/tests/Flow/Parquet/Tests/Unit/ParquetFile/Codec/ZSTDTest.php`

and make it specific, because the goal is to execute only those tests with `--testsuite=extension-zstd`

flavioheleno: On a second note, shall I group the tests (using [test-groups](https://docs.phpunit.de/en/10.5/attributes.html#test-groups)) so that only the specific tests are executed?

norberttech: > On a second note, shall I group the tests (using [test-groups](https://docs.phpunit.de/en/10.5/attributes.html#test-groups)) so that only the specific tests are executed?

Yes! That's much better, I totally forgot about test-groups, nice catch!

src/lib/parquet/src/Flow/Parquet/stubs.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
norberttech: I think I would move that file to `src/lib/parquet/src` since it's not under any namespace and `functions.php` reflects Flow\Parquet namespace

flavioheleno: I put it there initially, but then I saw `functions.php` and thought it was better to put them together :)) I'll move it around, no worries!

.github/workflows/test-groups.yml
```diff
@@ -0,0 +1,91 @@
+name: Test Groups that require Extensions
```
norberttech: Extensions Tests maybe? with `test-extensions` ?

src/lib/parquet/src/Flow/Parquet/ParquetFile/Codec.php
```diff
@@ -26,6 +26,7 @@ public function compress(string $data, Compressions $compression) : string
             Compressions::SNAPPY => \snappy_compress($data),
             /** @phpstan-ignore-next-line */
             Compressions::GZIP => \gzencode($data, $this->options->get(Option::GZIP_COMPRESSION_LEVEL)),
+            Compressions::ZSTD => \zstd_compress($data, $this->options->get(Option::ZSTD_COMPRESSION_LEVEL)),
```
flavioheleno: may I add an int typecast here to avoid the static analysis error?

norberttech: I believe there should be an `getInt` getter, and if not I would add one that would return ints so we wont need to cast anything anywhere :)

norberttech:
that was quick, thank you @flavioheleno 🍻

Issue #1124, opened 2024-07-20
Use a generic unique values generators across the whole project


norberttech:
Just like we are now going to use Hash\Algorithm across whole project thanks to https://github.com/flow-php/flow/pull/1122 by @mleczakm I think we could do the same with Unique values. 

Recently @stloyd replaced all `uniquid()` with `bin2hex(random_bytes(...))` in https://github.com/flow-php/flow/pull/1102 but I think we should take it to another level and create a `UniqueFactory` with following methods: 

```php
UniqueFactory::string(int $length) : string
UniqueFactory::int(int $start, int $end) : int
```

and that should be enough to replace most of our current use cases where we are using `bin2hex(random_bytes)`

PR #1128, opened 2024-07-20
Add UniqueFactory for creating random string or int from given range

mleczakm:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
     <li>UniqueFactory as source of random ints/strings with given length</li> 
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Solves https://github.com/flow-php/flow/issues/1124

### DIFF
--- a/src/core/etl/src/Flow/ETL/UniqueFactory.php
+++ b/src/core/etl/src/Flow/ETL/UniqueFactory.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL;
+
+class UniqueFactory
+{
+    public static function int(int $min, int $max) : int
+    {
+        return \random_int($min, $max);
+    }
+
+    /** @param int<1, max> $int */
+    public static function string(int $int) : string
+    {
+        $bytes = (int) \ceil($int / 2);
+        $bytes >= 1 ?: $bytes = 1;
+
+        return \mb_substr(\bin2hex(\random_bytes($bytes)), 0, \max(0, $int));
+    }
+}
--- a/src/core/etl/src/Flow/ETL/UniqueFactory.php
+++ b/src/core/etl/src/Flow/ETL/UniqueFactory.php
@@ -4,8 +4,12 @@ declare(strict_types=1);
 
 namespace Flow\ETL;
 
-class UniqueFactory
+final class UniqueFactory
 {
+    private function __construct()
+    {
+    }
+
     public static function int(int $min, int $max) : int
     {
         return \random_int($min, $max);
@@ -17,6 +21,6 @@ class UniqueFactory
         $bytes = (int) \ceil($int / 2);
         $bytes >= 1 ?: $bytes = 1;
 
-        return \mb_substr(\bin2hex(\random_bytes($bytes)), 0, \max(0, $int));
+        return \substr(\bin2hex(\random_bytes($bytes)), 0, \max(0, $int));
     }
 }
--- a/src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/PartitioningTest.php
+++ b/src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/PartitioningTest.php
@@ -163,13 +163,13 @@ final class PartitioningTest extends IntegrationTestCase
                     function (int $i) : array {
                         $data = [];
 
-                        $maxItems = \random_int(2, 10);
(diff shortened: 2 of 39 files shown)


src/core/etl/src/Flow/ETL/UniqueFactory.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL;
+
+class UniqueFactory
```
stloyd:
```suggestion
final class UniqueFactory
```

src/core/etl/src/Flow/ETL/UniqueFactory.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public static function int(int $min, int $max) : int
```
stloyd: Maybe also:
```suggestion
    private function __construct()
    {}

    public static function int(int $min, int $max) : int
```
Cause this class has only static methods.

src/core/etl/src/Flow/ETL/UniqueFactory.php
```diff
@@ -0,0 +1,22 @@
...
+        return \random_int($min, $max);
+    }
+
+    /** @param int<1, max> $int */
+    public static function string(int $int) : string
+    {
+        $bytes = (int) \ceil($int / 2);
+        $bytes >= 1 ?: $bytes = 1;
+
+        return \mb_substr(\bin2hex(\random_bytes($bytes)), 0, \max(0, $int));
```
stloyd: I think we don't need multibyte here, do we?

src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/PartitioningTest.php
```diff
@@ -163,13 +163,13 @@ public function test_partition_by_partitions_order() : void
                     function (int $i) : array {
                         $data = [];
 
-                        $maxItems = \random_int(2, 10);
```
stloyd, line 166: More I look at this, more I think I would add new function for unique int/string to "hide" object call & improve DX perspective, WDYT?

norberttech:
The PR look good and we could merge it like that (beside few small things @stloyd pointed out) but I think we might be able to still improve even more. Here are my thoughts. 

**Naming** - now when I look at it something is missing in the name, like `RandomValueGenerator` would be more descriptive. 

**Dependency Injection** - I think I would turn `RandomValueGenerator` into an interface with `NativePHPRandomValueGenerator` that we can simply inject into objects that would need a unique values (I'm thinking about caching strategies here and more advanced data processing algorithms). 

**Scalar Functions** - Additionally (can go to a standalone PR) we could also create Flow scalar functions that could be used during data processing, like: 

```php
df()
   ->read(...) 
   ->withEntry('random_string', random_string(length: 16))
   ->write(to_output())
   ->run()

df()
   ->read(...) 
   ->withEntry('random_string', random_string(length: ref('some_int_column'))))
   ->write(to_output())
   ->run()
```

This function would implement `Flow\ETL\Function\ScalarFunction` and could take `$lenght` as another `ScalarFunction` so this way length could be taken from a column or output of chained functions (some calculations). Additionally it would also need `RandomValueGenerator $generator = new NatievPHPRandomValueGenerator()` in case anyone would like to change the generation strategy. 

**DSL Functions** - as pointed by @stloyd we probably should also add `generate_random_string(int $length = 16)` and `generate_random_int(int $start = PHP_INT_MIN, int $end = PHP_INT_MAX, RandomValueGenerator $generator = new NatievPHPRandomValueGenerator())` functions so we can use them in tests instead of static factories. 

Thoughts?

src/core/etl/src/Flow/ETL/UniqueFactory.php
mleczakm:
```suggestion
        return \substr(\bin2hex(\random_bytes($bytes)), 0, \max(0, $int));
```

norberttech (changes requested):
Some changes described above are required to merge this PR :)

norberttech (approved):

PR #1149, opened 2024-07-28
Add `array_to_row` helper function

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `array_to_row` helper function</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/core/etl/src/Flow/ETL/DSL/functions.php
+++ b/src/core/etl/src/Flow/ETL/DSL/functions.php
@@ -796,8 +796,55 @@ function number_format(ScalarFunction $function, ?ScalarFunction $decimals = nul
 }
 
 /**
- * @psalm-suppress PossiblyInvalidIterator
- *
+ * @param array<array<mixed>>|array<mixed|string> $data
+ * @param array<Partition>|Partitions $partitions
+ */
+function array_to_row(array $data, EntryFactory $entryFactory = new NativeEntryFactory(), array|Partitions $partitions = [], ?Schema $schema = null) : Row
+{
+    foreach ($data as $key => $v) {
+        if (!\is_string($key)) {
+            throw new InvalidArgumentException('Passed array keys must be a string. Maybe consider using "array_to_rows()" function?');
+        }
+    }
+
+    $entries = [];
+
+    foreach ($data as $key => $value) {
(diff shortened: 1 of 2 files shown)


src/core/etl/tests/Flow/ETL/Tests/Unit/RowsTest.php
```diff
@@ -285,6 +285,65 @@ public function test_building_rows_from_array_with_schema_but_entries_not_availa
         );
     }
 
+    public function test_building_single_row_from_array() : void
+    {
+        $row = array_to_row(
+            [
+                ['id' => 1234, 'deleted' => false, 'phase' => null],
+                ['id' => 4321, 'deleted' => true, 'phase' => 'launch'],
+            ]
```
norberttech, line 297: hmm I'm honestly not sure how this edge case should be handled, because technically speaking this is a use case for `array_to_rows()` and the output that returns only one row when input defines two of them is a bit hard to predict. 

I'm tempted to say that in this case array_to_row should throw an exception. 

The validation could simply check if all array keys are strings (since only strings can become a entry names)

src/core/etl/src/Flow/ETL/DSL/functions.php
```diff
@@ -795,6 +795,78 @@ function number_format(ScalarFunction $function, ?ScalarFunction $decimals = nul
     return new NumberFormat($function, $decimals, $decimalSeparator, $thousandsSeparator);
 }
 
+/**
+ * @psalm-suppress PossiblyInvalidIterator
+ *
+ * @param array<array<mixed>>|array<mixed|string> $data
+ */
+function array_to_row(array $data, EntryFactory $entryFactory = new NativeEntryFactory(), ?Schema $schema = null) : Row
```
norberttech: once this is implemented it can be used in array_to_rows I believe while iterating over rows to reduce the code duplication

PR #1151, opened 2024-07-28
Add XML writer

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add XML writer</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
Just to give some context, @stloyd created this PR long time ago but back then I decided to not merge it because it was missing two things: 

- handling deeply nested data structures lists/maps/structs
- handling attributes

The first one is a similar concept that we are using in ParquetConverter - recursive iteration over Flow entries and converting them into XML
The second one can be implemented based on naming convention, for example, if a entry has `__xml_attribute__{entry_name}` prefix (or something different) it would be added as an attribute to row node (or struct/list/map). A helper function would be nice, something like `entry_to_xml_attribute(ref(...))`. 

Besides those two things, it is ready to merge. If anyone wants to take it over (not sure when I will have time, XML is not the highest priority for me right now but it's definitely important) I would kindly ask to pull the code directly from Joseph's branch so we can keep the clean history of contributions.

Issue #745, opened 2023-11-06
Add PHPUnit internal helper(s)


stloyd:
We often start to do copy-paste work in tests, thus we should begin to have some reusable helpers to make our internal tests easier to maintain.

Helpers:
- `FilesystemHelper` with: `create`, `delete` for files & directories,
- ...

Bellangelo:
Hi @norberttech , I would like to help on this. If I am not mistaken there isn't any generic place to create the new helpers. Should we create a `tests` directory at the root for this case?

norberttech:
hey @Bellangelo, awesome! I think I would put it inside of the `src/core/etl/src/Flow/ETL/Test/...` so it can be installed with ETL package together

Bellangelo:
@norberttech, i checked and the Filesystem issue that the original PR was addressing seems to be fixed. I checked a little bit around the project and I noticed that some of the Extractor's tests logic is duplicated so I created a PR for it https://github.com/flow-php/flow/pull/1291. Not sure if all the changes are needed but it is a start.

Also, since the description of this issue does not contain much info I would really appreciate it if you could add what it is missing currently.

norberttech:
Since this task was never properly explained, here is a proper description of what we are trying to achieve here: 

https://github.com/flow-php/flow/pull/1291#issuecomment-2564576952

norberttech:
I think we can create some very generic fake extractors and define schema for them,for example

- df()->read(from_flow_orders(limit:1_000))
- flow_orders_schema() : Schema
- do the same for products
- customers
- inventory

and we should make sure all of those datasets keep consistent schema and that they are using all possible entry types.
Those virtual datasets would need to follow
very strict backward compatibility policy and proper schema evolution. 

This would make so much easier and more realistic to test not only entire pipeline but also stand-alone scalar functions as we can also create helpers that would just give us one row (and use them inside fake extractors)

PR #1291, opened 2024-12-28
Extractor test helper

Bellangelo:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>FlowTestCase with assertions around Extractors</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Partially addresses https://github.com/flow-php/flow/issues/745.

I found some duplicated logic in the Extractor's tests and I decided to unify the logic around them.


src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/ExtractorTestCase.php
```diff
@@ -0,0 +1,50 @@
...
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Tests\Unit\Extractor;
+
+use Flow\ETL\{Config, Extractor, FlowContext, Row};
+use PHPUnit\Framework\TestCase;
+use function iterator_to_array;
+
+abstract class ExtractorTestCase extends TestCase
```
Bellangelo: If we decide that these methods do indeed help maybe it would a better idea to convert this abstract class into a trait or something similar. The reason is that there are cases outside the Extractors tests where we counting or converting the rows.

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/ExtractorTestCase.php
```diff
@@ -0,0 +1,50 @@
...
+            $totalRows += $rows->count();
+        }
+
+        static::assertSame($expectedTotalCount, $totalRows);
+    }
+
+    public function assertCountRows(int $expectedCount, Extractor $extractor) : void
+    {
+        static::assertCount($expectedCount, $this->toRowsArray($extractor));
+    }
```
Bellangelo: Maybe these 2 should be a single method?

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/ExtractorTestCase.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Flow\ETL\Tests\Unit\Extractor;
+
+use Flow\ETL\{Config, Extractor, FlowContext, Row};
+use PHPUnit\Framework\TestCase;
+use function iterator_to_array;
+
+abstract class ExtractorTestCase extends TestCase
+{
+    public function assertCountMultiRows(int $expectedTotalCount, int $expectedCountPerRow, Extractor $extractor) : void
```
Bellangelo: I am not sure if the `int $expectedCountPerRow` should be instead a callback. The reason is that there are cases where we asserting the schema instead. For example, this is happening in [JsonExtractorTest](https://github.com/flow-php/flow/blob/1.x/src/adapter/etl-adapter-json/tests/Flow/ETL/Adapter/JSON/Tests/Integration/JSONMachine/JsonExtractorTest.php#L112).

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/RowsExtractorTest.php
```diff
@@ -36,7 +29,7 @@ public function test_process_extractor() : void
                 ['number' => 4, 'name' => 'four'],
                 ['number' => 5, 'name' => 'five'],
             ],
-            $rows->toArray()
```
Bellangelo, line 39: We had a bug here. It should be using the `$data` variable.

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/ExtractorTestCase.php
norberttech: I'm not sure if I understand what this function is checking, it's collecting the total number of rows from all processed batches?

Bellangelo: @norberttech Hmm.. I think you are right. I might have done a blind copy/paste here without thinking about it much. in this case the name of the method should change.

Maybe `assertTotalBatchesRows`?

norberttech: since this is extractor-related I would probably go with something like: 

`assertExtractorYieldedRows(int $expectedRows, Extractor $extract, string $message = null)`

I think it would improve the developer experience if we group those assertions around what they are doing, like this: 

```
self::assertExtractorXXX
self::assertLoaderXXX
...
```

norberttech:
hey @Bellangelo it's a great start! 
After looking at your proposal few things comes to my mind. 

- focus that helper class on assertions (no helper methods)
- name it `FlowTestCase`
- it should be a part of the `src/core/etl`
- I would try to put it under `src/core/etl/src/Flow/ETL/Test` namespace to make it available with the ETL 

and ideally, I would probably try to make all Flow test cases to extend from it, at least across those that depends on `flow-php/etl`

norberttech:
also @Bellangelo feel free to start a thread at [discord](https://discord.com/channels/1263821077558333451/1264131793616113674) if you have any questions about this task. 
We can talk in more details about it there if you prefer more live time communication

Bellangelo:
@norberttech I think I have finished the method changes. The only thing I haven't done is this one:
> * I would try to put it under `src/core/etl/src/Flow/ETL/Test` namespace to make it available with the ETL

Is the namespace correct or you maybe meant to write `src/core/etl/tests/Flow/ETL/Tests`? The reason I am asking is that it concerns me that we will have a class in the "main" src that depends in a dev dependency, the PHPUnit. Maybe I am missing something?

norberttech:
@Bellangelo, thank you for your work, and I'm sorry for the lack of proper guidance. Let me explain the vision once again and update the issue. 

So in general my goal was to provide a FlowTestCase class that people could use to test their pipelines. 
I would focus this one on: 
What you started to build here is more of a candidate for the internal use cases (because regular users should not write their extractors/loaders). 

We already have [IntegrationTestCase](https://github.com/flow-php/flow/blob/0407556d33a12576772feff203764bc15cdfe202/src/core/etl/tests/Flow/ETL/Tests/Integration/IntegrationTestCase.php#L13) which is doing something similar to what you are trying to achieve here.

We can approach it in a following way : 

InternalTestCase (those should sit under `src/core/etl/tests/Flow/ETL/Tests`)

 - FlowTestCase - the one that you started building, should be used with all Unit Tests 
 - FlowIntegrationTestCase that extends FlowTestCase - this could be the one we already have under IntegrationTestCase 

ExternalTestCase 

- FlowTestCase 

This one should be exposed with the ETL itself or maybe we can put it under a dedicated package like `flow-php/bridge-phpunit` with some dedicated helpers that TestCases for things like ScalarFunctions for example.

So what you did so far is a good start to unify InternalTestCases across the whole repository and if you don't mind I would like you to continue with this one.

I'm going to leave some comments in the code directly. 

Please let me know if that makes sense and again, feel free to reach out on Discord if you prefer a more synchronous way of communication

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
```diff
@@ -0,0 +1,71 @@
...
+        if (!$extractorContainsBatches) {
+            static::fail('Extractor does not contain any batches');
+        }
+    }
+
+    /**
+     * @param array<Rows> $expectedRows
+     * @param Extractor $extractor
+     */
+    public function assertExtractorEqualsRows(array $expectedRows, Extractor $extractor) : void
```
norberttech: maybe `assertExtractedRowsEquals(Rows $rows, Extractor $extractor, ?string $message = null) : void`

and under the hood what you can do is merge all the rows extracted by the extractor:

```php
$extractedRows = rows();

foreach ($extractor->extract(flow_context()) as $nextRows) {
    $extractedRows = $extractedRows->merge($nextRows);
}

self::assertEquals($expextedRows, $extractedRows);
```

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
```diff
@@ -0,0 +1,71 @@
...
+declare(strict_types=1);
+
+namespace Flow\ETL\Tests\Unit\Extractor;
+
+use Flow\ETL\{Config, Extractor, FlowContext, Rows};
+use PHPUnit\Framework\TestCase;
+
+abstract class FlowTestCase extends TestCase
+{
+    public function assertExtractorCountBatches(int $expectedCount, Extractor $extractor) : void
```
norberttech: I think we should make all the methods static and final

norberttech: additionally, to all methods that need to execute Extractor, I would add an extra options argument for FlowContext so in case it's needed it can be configured in some way

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
norberttech: and I would then make a similar method `assertExtractedRowsAsArrayEquals`

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        static::assertCount(
+            $expectedCount,
+            \iterator_to_array($extractor->extract(new FlowContext(Config::default())))
+        );
+    }
+
+    public function assertExtractorCountRows(int $expectedCount, Extractor $extractor) : void
```
norberttech: `assertExtractedRowsCount`

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
norberttech: `assertExtractedBatchesCount`

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
```diff
@@ -0,0 +1,71 @@
...
+        $totalRows = 0;
+
+        foreach ($extractor->extract(new FlowContext(Config::default())) as $rows) {
+            $totalRows += $rows->count();
+        }
+
+        static::assertSame($expectedCount, $totalRows);
+    }
+
+    public function assertExtractorCountRowsPerBatch(int $expectedCount, Extractor $extractor) : void
```
norberttech: `assertExtractedBatchesSize`

Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1291/commits/87f5a00d47f4a0842286506070275c55b786f64f

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1291/commits/87f5a00d47f4a0842286506070275c55b786f64f

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
Bellangelo: > assertExtractedBatchesCount

Fixed in assertExtractedBatchesCount

> additionally, to all methods that need to execute Extractor, I would add an extra options argument for FlowContext so in case it's needed it can be configured in some way

Fixed in https://github.com/flow-php/flow/pull/1291/commits/7294309d14c409368408f34cd80e132f3e9f1271

> I think we should make all the methods static and final

Fixed in https://github.com/flow-php/flow/pull/1291/commits/466151b9ccdd87ddce3c55305370b51fcf9eb8f8

src/core/etl/tests/Flow/ETL/Tests/Unit/Extractor/FlowTestCase.php
Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1291/commits/315d9f96f0dbb6bd19031ea43392d183f53a4173 and https://github.com/flow-php/flow/pull/1291/commits/edef63a70f5c86bbeaa1a197a3439077ffc83390

src/core/etl/tests/Flow/ETL/Tests/FlowTestCase.php
```diff
@@ -0,0 +1,94 @@
...
+
+abstract class FlowTestCase extends TestCase
+{
+    final public static function assertExtractedBatchesCount(
+        int $expectedCount,
+        Extractor $extractor,
+        ?FlowContext $flowContext = null,
+        string $message = '',
+    ) : void {
+        $flowContext = $flowContext ?? new FlowContext(Config::default());
```
norberttech: could you please use DSL functions in test suite instead of objects? 

```php
$flowContext = $flowContext ?? flow_context();
```

Bellangelo:
> @Bellangelo, thank you for your work, and I'm sorry for the lack of proper guidance.

I guess we are even due to how bad I am consistently naming these assertions 😃 .


> InternalTestCase (those should sit under `src/core/etl/tests/Flow/ETL/Tests`)
> 
> * FlowTestCase - the one that you started building, should be used with all Unit Tests
> * FlowIntegrationTestCase that extends FlowTestCase - this could be the one we already have under IntegrationTestCase

Done in https://github.com/flow-php/flow/pull/1291/commits/1e1c2feaa7103666e94101f65e4cb2f0691a540d and https://github.com/flow-php/flow/pull/1291/commits/6b300a0ddf179a67e70c702a48ee339a52db055e. I haven't updated all the tests to extend the new `FlowTestCase` since it isn't required by all the classes. Plus, it is a 100+ file changes if I do this. Although if you think we should do it, I can happily open a new PR for it after we merge this one.

src/core/etl/tests/Flow/ETL/Tests/FlowTestCase.php
Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1291/commits/cea9939f2fe2388a04cdf38687d18f77f7552da2

PR #1329, opened 2025-01-05
Run PHPStan for the test suite

### LINKED ISSUE #1340: Run PHPStan for the test suite
Currently, we run PHPStan only for our main code. However, tests are also an important part of our codebase and we need to ensure their quality as well.

Bellangelo:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Run PHPStan for test suite</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes: https://github.com/flow-php/flow/issues/1340


Bellangelo:
Hi @norberttech, codecov failed for some reason and it points to `lib/snappy` not being fully covered but I haven't touched this file. Do you have any idea what the bug might be here?

norberttech:
don't worry about snappy, I think tests there are a bit fragile. 
How hard it would be to fix those issues instead of using baseline? I’m not sure how much I like merging it, that's the reason why in the pr about hardening phpstan generics analysis I fixed around 350~ issues which took me far less than I expected to be honest so maybe it's worth investing some extra time and get rid of baseline?

.github/workflows/static-analyze-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+    paths-ignore:
+      - 'CHANGELOG.md'
+
+# See https://stackoverflow.com/a/72408109
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  static-analyze:
```
norberttech: this can be merged with the other static analysis workflow file, too many workflow files is not great

.github/workflows/static-analyze.yml
```diff
@@ -95,4 +95,4 @@ jobs:
         run: "composer static:analyze:cs-fixer"
 
       - name: "Static Analyze - PHPStan"
-        run: "composer static:analyze:phpstan -- --error-format=github"
+        run: "composer static:analyze:phpstan --error-format=github"
```
norberttech: I think this is incorrect, without double — composer is not passing anyoption to phpstan

norberttech:
question, why do we need a different phpstan configuration for test suite?

.github/workflows/static-analyze.yml
Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1329/commits/7a4859e6c1400ad078c8b248106ca19331548e9c

.github/workflows/static-analyze-tests.yml
Bellangelo: Fixed in https://github.com/flow-php/flow/pull/1329/commits/6f4bf08ddc59b6178e68dc4367086b0c21c4f295

Bellangelo:
> question, why do we need a different phpstan configuration for test suite?

Splitting them apart would allow us to have different quality checks for each case. For example, we could have a lower level in our test suite or skip specific errors. Although now, i have unified the configs in this commit https://github.com/flow-php/flow/pull/1329/commits/6f4bf08ddc59b6178e68dc4367086b0c21c4f295

Bellangelo:
Hi @norberttech, we have these ignorred errors in our PHPStan config:

> ignoreErrors:
        - identifier: argument.type
        - identifier: missingType.iterableValue

I think we should ignore them using the baseline. The reason is that these errors are not specific to any files or number of occurrence which means that we can keep adding these errors and PHPStan will happily ignore them.

norberttech:
> Hi @norberttech, we have these ignorred errors in our PHPStan config:
> 
> > ignoreErrors:
> > - identifier: argument.type
> > - identifier: missingType.iterableValue
> 
> I think we should ignore them using the baseline. The reason is that these errors are not specific to any files or number of occurrence which means that we can keep adding these errors and PHPStan will happily ignore them.

Based on our conversation at Discord I created an ADR that answers this one #1345

Bellangelo:
@norberttech All errors have been fixed. It took a while but no baseline has been used 🚀

norberttech:
you are THE BEST 💪 will check and merge it tomorrow morning

PR #1330, opened 2025-01-05
Enforce usage of DSL functions in test suite

### LINKED ISSUE #1301: Enforce usage of DSL functions in test suite
In test suite we should relay mostly on DSL functions instead of an object-oriented approach. 
One of the possibilities is to use PHPStan for that. I would rather like to avoid adding new tools and also using Psalm that I'm thinking about removing from the project

Bellangelo:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Enforce the usage of DSL functions in tests</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>
Parent: https://github.com/flow-php/flow/pull/1329

Closes https://github.com/flow-php/flow/issues/1301

Used `PHPParser` to traverse the `functions.php` and print the desired config. This script is missing from the PR since it could be only a one-time thing.


Bellangelo:
@norberttech The total number of errors is [2189](https://github.com/flow-php/flow/actions/runs/12623079283/job/35171684313?pr=1330). This shouldn't be merged yet as some of them could be valid cases which I haven't confirmed yet.
Before we continue with any changes, is this the desired behavior?

norberttech:
maybe instead of parsing functions.php we could start from not allowing to initialize certain objectets in test suite? 

- Entry
- Type
- Definition
- Schema
- Row
- Rows 
- Scalar Function
- Window Function

Those are probably 90% (maybe more) of total cases

Bellangelo:
It was decided in a private discussion in Discord that we will use Rector for this use case.

PR #1345, opened 2025-01-08
Introduce ADR's to the project

### LINKED ISSUE #1344: Start writing down ADR's
With new contributors joining the project, it might be difficult to explain all the historical decisions we made over time. 
It's also not easy to remember all of those decisions over time, at this point, Flow has 26 sub-repositories. 
To solve that problem all major architectural decisions should be described in ADR documents that every single contributor **must** follow with no exceptions. 
The only way to merge a code that is against an existing ADR is to write a new ADR that overrides the previous one.

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Project ADR's</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Resolves: #1344 

Since this is strongly impacting contributions I would love to hear opinions about this PR. 

@stloyd 
@mleczakm 
@Bellangelo 

I mentioned only the most active recent contributors to approve that PR but anyone is welcome to participate in this conversation, we care about all opinions 🤗 

Of Course, ADR's are going to be available also at https://flow-php.com which thanks to amazing https://github.com/thephpleague/commonmark library allowed us to also have "mentions" and many more amazing features!

### DIFF
--- a/documentation/adrs/static-analysis-baseline.md
+++ b/documentation/adrs/static-analysis-baseline.md
@@ -1,0 +1,50 @@
+# Static Analysis Baseline
+
+Proposed by: @norberttech
+Date: 2025-01-07
+
+## Context
+---
+
+After removal of Psalm and making PHPStan the main static analysis tool, 
+we looked again into hardening the static analysis configuration.
+
+We had three globally ignored errors in phpstan configuration
+
+- identifier: argument.type
+- identifier: missingType.iterableValue
+- identifier: missingType.generics
+
+All of the above were significantly reducing the value of static analysis and 
+code quality.
+
+One of the proposed approaches was to use a baseline file to suppress those errors
+in the existing codebase and gradually remove them.
+
+There are few problems with this approach, but the most significant one is that once the baseline is introduced, 
+it needs to be maintained.  
+Whenever the code is changed, the baseline needs to be updated, which is an opportunity to also suppress new errors.  
+This means that maintainers would not only need to go through the code changes
+but also through the baseline file, which is not the best use of their very limited and valuable time anyway.
+
+## Decision
+---
+
+We **must not** use baseline for static analysis. 
+Instead, errors can be suppressed by annotations in the codebase or globally
+in the static analysis tool configuration.
+
+Error suppression should be considered an edge case and should be used sparingly.
+Core contributors should review and approve all suppression annotations.
+
+## Consequences
+---
+
+- The codebase will be cleaner and more maintainable.
+- More predictable and stricter types.
+- Less maintenance overhead related to managing the baseline.
+
+## Links and References
+---
+
+- #1329
--- a/documentation/adrs/static-analysis-baseline.md
+++ b/documentation/adrs/static-analysis-baseline.md
@@ -37,7 +37,7 @@ in the static analysis tool configuration.
 Error suppression should be considered an edge case and should be used sparingly.
 Core contributors should review and approve all suppression annotations.
 
-## Consequences
+## Pros & Cons
 ---
 
 - The codebase will be cleaner and more maintainable.
--- a/documentation/adrs/template.md
+++ b/documentation/adrs/template.md
@@ -20,7 +20,7 @@ Date: YYYY-MM-DD
 > What is the decision being made?  
 > State it clearly and concisely.
 
-## Consequences
+## Pros & Cons
 ---
 
 `// Todo: Describe the consequences here`
--- a/documentation/adrs/template.md
+++ b/documentation/adrs/template.md
@@ -1,0 +1,43 @@
+# [Decision Title]
+
+Proposed by: @author_github_username
+Date: YYYY-MM-DD
+
+## Context
+---
+
+`// Todo: Describe context here`
+
+> What is the current situation or issue that needs addressing?  
+> Why is this decision necessary?  
+> Any relevant background or constraints.
+
+## Decision
+---
+
+`// Todo: Describe the decision here`
+
+> What is the decision being made?  
+> State it clearly and concisely.
+
+## Consequences
+---
+
+`// Todo: Describe the consequences here`
+
+> What are the outcomes or implications of this decision?  
+> Highlight both positive and negative impacts.
+
+
+## Alternatives Considered (optional)
+---
+
+`// Todo: Describe the alternatives here`
+
+> Briefly describe any alternatives that were considered.  
+> Why were they not chosen?
+
+## Links and References (optional)
+---
+
+> Include any supporting documents, links to issues, pull requests, or discussions that provide context.
--- a/documentation/contributing.md
+++ b/documentation/contributing.md
@@ -6,6 +6,11 @@ Even that we are supporting 3 PHP versions at time, we are using the lowest supp
 
 For the code coverage, please install [pcov](https://pecl.php.net/package/pcov).
 
+### Before you change anything
+
+Please make sure that you are aware of our [Architecture Decision Records](/documentation/adrs.md).
+It's mandatory to follow all of them without any exceptions unless explicitly overridden by a new ADR.
+
 ### Prepare Project:
 
 ```shell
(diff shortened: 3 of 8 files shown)


documentation/adrs/template.md
```diff
@@ -0,0 +1,43 @@
+# [Decision Title]
```
Bellangelo, line 1: What template is this one? It look close to a [MADR](https://github.com/adr/madr/blob/4.0.0/template/adr-template-minimal.md) template.

norberttech: Norbert template 🤣 I just put that together based on what worked for me before at bigger scale than we are

documentation/contributing.md
```diff
@@ -26,6 +26,11 @@ composer test
 composer static:analyze
 ```
 
+### Before you change anything
+
+Please make sure that you are aware of our [Architecture Decision Records](/documentation/adrs.md).
+It's mandatory to follow all of them without any exceptions unless explicitly overridden by a new ADR.
```
stloyd: This should be at top of this file, in first paragraphs IMO.

documentation/adrs/static-analysis-baseline.md
```diff
@@ -0,0 +1,50 @@
+# Static Analysis Baseline
+
+Proposed by: @norberttech
+Date: 2025-01-07
```
stloyd, line 4: What I miss is: `Status: proposal/accepted/denied` etc.

documentation/adrs/template.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Decision
+---
+
+`// Todo: Describe the decision here`
+
+> What is the decision being made?  
+> State it clearly and concisely.
+
+## Consequences
```
stloyd: IMO more useful can be pros & cons, especially if there is more than one approach proposed

documentation/adrs/template.md
Bellangelo: Based on the comments of @stloyd, maybe we could use the MADR template? You can find a more complete version of it here: https://github.com/adr/madr/blob/4.0.0/template/adr-template.md

There are also a lot of tools about it: https://adr.github.io/adr-tooling/

norberttech: sure we can, but like with everything, I need some reason 😅 Do you have some past experience with using that template? How it worked over time? Any sections become problematic at some point? Did anyone actually contributed?
Or maybe it was perfect and you would nit change it at all? 

I'm not trying here to reject that idea, but I have this principal that whenever alternatives are proposed they must come with clear benefits, otherwise its just art for art's sake.

Anyway I'm open, what would be also helpful when proposing alternatives is to provide the same ADR that I wrote under a different template so we can easier compare them

Bellangelo: I get your points. Despite my personal experience with it, I mainly suggested it because your template is already close to its minimal version ( as per my last comments ) and adding what stloyd suggested ( https://github.com/flow-php/flow/pull/1345#discussion_r1907586621, https://github.com/flow-php/flow/pull/1345#discussion_r1907587509 ) basically makes it the full version which I posted in my last comment.

Back to my experience, we use it in my current job and we don't have any problems with it. At the moment only people in the architecture team write them.

I have created this [PR](https://github.com/flow-php/flow/pull/1349) which uses the MADR template for the ADR.

documentation/adrs/static-analysis-baseline.md
norberttech: There is an explanation already in the ADR's readme that any ADR that is merge is considered accepted, propals are open PR's, denied are closed PR - adding status here IMO is redundant. 

My goal with this template is to make it as minimalistic as possible and as descriptive as possible, that's why it does not come with all the sections that are mentioned by @Bellangelo

documentation/adrs/template.md
norberttech: > Despite my personal experience with it, I mainly suggested it because your template is already close to its minimal version ( as per my last comments ) and adding what stloyd suggested ( https://github.com/flow-php/flow/pull/1345#discussion_r1907586621, https://github.com/flow-php/flow/pull/1345#discussion_r1907587509 ) basically makes it the full version which I posted in my last comment.

yeah, but this is still an art for art's sake since it does not bring any additional benefits. 

For example those two sections: 

```md
## Decision Drivers

* Availability of maintainers
* Maintenance costs

## Considered Options

* No baseline file - keep everything as is
* Add errors to the baseline file
```

From my point of view, they are not making ADR better in any way, and the only thing they will create is deter people from contributing/writing ADR's (been there, done that). 

Splitting `Decision Outcome` into `Consequences` and `Confirmation` is also not making it better IMO, just more granular but that granularity is not useful anyway. 

> Back to my experience, we use it in my current job and we don't have any problems with it. At the moment only people in the architecture team write them.

That's one of the things I don't appreciate, some architects think that only architects can write ADR's and they deliver them to the team instead of building them together with the team. 
It's pretty much a side effect of making the ADR template overengineered. That approach might work for some systems, but IMO it's not in the spirit of the open source. 
I want to make ADR's minimalistic, but still precise, easy to understand but detailed enough to eliminate conflicts. I aim to allow anyone to write an ADR, not just an "architects".  

So long story short, unless you can provide some strong arguments about how the project will benefit from having those extra sections I would still move forward with the more minimalistic version.

@stloyd what's your opinion on this one?

norberttech:
Implemented changes suggested by @stloyd and added ADR about Extension Points in the project

documentation/adrs/static-analysis-baseline.md
stloyd: Alternatively for denied there could be list of PRs with specific tag and link to that list in ADR list, so it would be easily to find if someone already proposed something that got denied, wdyt?

Bellangelo (approved):

documentation/adrs/template.md
Bellangelo: > From my point of view, they are not making ADR better in any way, and the only thing they will create is deter people from contributing/writing ADR's (been there, done that).

The section provide a quick overview of the ADR without reading it completely. Its more like the summary of a book and that's why they are at the top and the details are in the bottom.

> Splitting Decision Outcome into Consequences and Confirmation

Here is just a separation of concerns. The `Decision` should be mostly about the chosen option. The `Consequences` are the good and the bad things of the chosen option and the `Confirmation` is how this option would be enforced / applied.

You could indeed group them together, as it happens in the minimal template of MADR.

documentation/adrs/static-analysis-baseline.md
norberttech: hmm we could put a label "ADR" and a link to all closed PRs with label ADR under index od merged ADRs as "Rejected ADRs"

documentation/adrs/template.md
norberttech: long story short, nothing in madr makes is superior to this one 😅 
look, at the end of the day, someone also put it together like this one here and its pretty much a matter of personal preferences which one to use. My preference is to keep it sharp and minimalistic, tailored precisely for our needs.
I'm gonna move forward with this template and if at any point in time it will turn out that something is missing it will get adjusted

documentation/adrs/static-analysis-baseline.md
stloyd: That's what I had in mind 👍

norberttech: added 🤝

PR #1394, opened 2025-01-19
Merge concat functions into one

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Merge concat functions into one</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
The reason why I did not merge those two is that this is aligned with SQL, which makes easier to write automated SQL -> Flow Transformations.

It also keeps them smaller and more predictable (less arguments changing function output).

I'm personally against merging them into one

stloyd:
I don't have hard feelings on merging them or not, for me strange was that internals are almost same, and the difference is that one has argument in front second not.

norberttech:
Totally, was strange for me too at first but after playing a bit more with SQL Parsers and potential support for pipelines defined in SQL it become more logical approach.

I'm gonna close this one for now, but might get back to it once I start working on that SQL support

PR #1405, opened 2025-01-26
Unified joins behavior

### LINKED ISSUE #1212: Unify join and joinEach behavior
Currently join and joinEach behaves a bit differently. 

join is using `HashJoin` algorithm under the hood when `joinEach` due is based on a nested loop algorithm. 

The problem is that the implementation of Nested Loop enforces using join_prefix because if we try to join two dataframes on `id` column when on both sides this column is called `id` we are going to get DuplicatedEntriesException coming from `Rows::merge()` method. 

What we should do is to remove from the right dataset join columns to avoid duplicates.

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Join columns from a join expression won't trigger the join duplicated columns exception anymore.</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Resolves: #1212

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,7 @@
         colors="true"
         displayDetailsOnTestsThatTriggerWarnings="true"
         displayDetailsOnTestsThatTriggerErrors="true"
+        cacheDirectory="var/phpunit/cache"
 >
   <php>
     <env name="AZURITE_HOST" value="localhost"/>
--- a/examples/topics/schema/apply/output.txt
+++ b/examples/topics/schema/apply/output.txt
@@ -1,4 +1,4 @@
 schema
 |-- id: integer
-|-- name: string
+|-- name: ?string
 |-- active: boolean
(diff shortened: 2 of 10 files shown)


phpunit.xml.dist
```diff
@@ -6,6 +6,7 @@
         colors="true"
         displayDetailsOnTestsThatTriggerWarnings="true"
         displayDetailsOnTestsThatTriggerErrors="true"
+        cacheDirectory="var/phpunit/cache"
```
norberttech, line 9: @Bellangelo So the result cache file option was removed, but the result cache is always stored in the PHPUnit cache folder or in the root directory. Adding this option moves the result cache to the PHPUnit cache folder.

PR #1407, opened 2025-01-26
Remove usage of `References::init` in `Row` class

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Remove usage of `References::init` in `Row` class</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Lower code works with both string & `Reference` so there is no point to enforce `Reference` class from string values.

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row.php
+++ b/src/core/etl/src/Flow/ETL/Row.php
@@ -6,7 +6,7 @@ namespace Flow\ETL;
 
 use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\Hash\{Algorithm, NativePHPHash};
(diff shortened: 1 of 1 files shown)


norberttech:
Makes sense!

PR #1440, opened 2025-02-02
Added composer.json to each example and make it standalone downloadable project

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>composer.json file to each example</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

image


.github/workflows/test-suite.yml
```diff
@@ -248,6 +248,7 @@ jobs:
         run: "composer test:adapter:xml"
 
       - name: "Test - Examples"
+      if: ${{ matrix.php-version == '8.2' && matrix.dependencies == 'locked' }}
```
norberttech: After releasing 0.11.0 this can be removed. Some examples are failing because if all dependencies are not explicitly provided like `array-dot` which a dependency of `etl` composer will not install dependencies due to `array-dot` php version constraint in release `0.10.0`

norberttech: (fixed indention)

PR #1447, opened 2025-02-05
Add JSONL read support

### LINKED ISSUE #1443: Support reading from datasets saved as jsonl
Currently JSONL is supported only when writing, for symmetry we must add it also to the reader. 

When reading jsonl all we need to do is read the json file line by line and, decode each line and turn it into a row that would be yielded after. 

DX should be similar to writter: 

`from_json()->asJsonl()`

jmortlock:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add read support for JSONL files</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Closes https://github.com/flow-php/flow/issues/1443

Support reading from JSONL based data sources.


src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonExtractor.php
```diff
@@ -91,6 +102,30 @@ public function withSchema(Schema $schema) : self
         return $this;
     }
 
+    private function getIterator()
```
norberttech: I would put a return type here `\Generator`

jmortlock:
For the tests I just replicated the existing ones, might be cleaner to use a DataProvider so there is not so much duplication, If you want I can switch it over.

As for the implementation I though it simpler to change the reading of the json lines to match what the original json behaviour is doing, that is it iterates out one complete object at a time. When use a pointer its assumed (like currently) that its pointing at an array subentry.

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonExtractor.php
jmortlock: Thats quick, i had that initially but get this type error

```
TypeError: Flow\ETL\Adapter\JSON\JSONMachine\JsonExtractor::getIterator(): Return value must be of type Generator, Closure returned
```
I can dig a bit deeper

jmortlock: I've changed it to return "callable" which i guess is correct as its returning a function that will generate an iterator. Maybe getIterator is not the best name

norberttech:
>For the tests I just replicated the existing ones, might be cleaner to use a DataProvider so there is not so much duplication, If you want I can switch it over.

I actually don't mind that duplication, 

> As for the implementation I though it simpler to change the reading of the json lines to match what the original json behaviour is doing, that is it iterates out one complete object at a time. When use a pointer its assumed (like currently) that its pointing at an array subentry.

From my point of view the pointer should be either skipped when we are working with jsonl or we can literally extract separated JsonLinesExtractor and JsonLinesLoader. 

The only way to make pointer works here from my point of view is to still read jsonl line by line and then apply pointer on each line. However JsonMachine library that we are using under the hood does not seem to support jsonl (but I just briefly checked their readme) we might be better making jsonl without support for pointers (also they are pretty much solving the same problem just in a different way). 

So the question is, how do you feel about extracting Json Lines logic to new classes?

It would be something like this: 

- JsonLinesExtractor - `from_json_lines()` 
- JsonLinesLoader - `to_json_lines()`

just without the pointer option on JsonLinesExtractor.

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonExtractor.php
norberttech: I think that if we split JsonLines and regular Json extractors/loaders this problem might even go away

jmortlock: Yes I did think that might be worthwhile as well, having a seperate JsonLExtractor and then using ```from_jsonl``` in the DSL.

```
                schema: $schema = df()
                    ->read(from_jsonl(__DIR__ . '/../../Fixtures/timezones.jsonl')))
```

I think JSONL still needs two iterators one for when its pointing at an Array Sub-entry and the case where there is no pointer and its reading as an Object; the iterator from JSONMachine behaves very differently in these cases.

jmortlock:
> > For the tests I just replicated the existing ones, might be cleaner to use a DataProvider so there is not so much duplication, If you want I can switch it over.
> 
> I actually don't mind that duplication,
> 
> > As for the implementation I though it simpler to change the reading of the json lines to match what the original json behaviour is doing, that is it iterates out one complete object at a time. When use a pointer its assumed (like currently) that its pointing at an array subentry.
> 
> From my point of view the pointer should be either skipped when we are working with jsonl or we can literally extract separated JsonLinesExtractor and JsonLinesLoader.
> 
> The only way to make pointer works here from my point of view is to still read jsonl line by line and then apply pointer on each line. However JsonMachine library that we are using under the hood does not seem to support jsonl (but I just briefly checked their readme) we might be better making jsonl without support for pointers (also they are pretty much solving the same problem just in a different way).
> 
> So the question is, how do you feel about extracting Json Lines logic to new classes?
> 
> It would be something like this:
> 
>     * JsonLinesExtractor - `from_json_lines()`
> 
>     * JsonLinesLoader - `to_json_lines()`
> 
> 
> just without the pointer option on JsonLinesExtractor.

Yup I like the sounds of that, and in my head was thinking along the same lines. 
Ill push up an alternate PR with your recommendations and we can go from there.

jmortlock:
Actually I do remember why i though pointer might still be useful for JSONL is that it could still save one level of transformation. For example say you have a JSONL file

{ data: [], pagination: [] }
{ data: [], pagination: [] }
{ data: [], pagination: [] }

You might just want to transform the "data" elements into your dataframe rows.

The main difference from a JSONMachine point of view is the iterating over object vs array as mentioned above.

norberttech:
> Ill push up an alternate PR with your recommendations and we can go from there.

You can even push it directly here, it would help us to keep the conversation history in one place in case anyone would like to know why we decided to separate them 😁

> You might just want to transform the "data" elements into your dataframe rows.
> 
> The main difference from a JSONMachine point of view is the iterating over object vs array as mentioned above.

Yeah, that's a good point, we can keep it but apply it independently on each row

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonExtractor.php
norberttech: > I think JSONL still needs two iterators one for when its pointing at an Array Sub-entry and the case where there is no pointer and its reading as an Object; the iterator from JSONMachine behaves very differently in these cases.

hmm would it be possible to just execute that callable and return the result which would be a generator (iterable)?

jmortlock:
Alright I've separated out the JSONLExtractor see what you think, once your happy ill do the loader separation.

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonlExtractor.php
```diff
@@ -0,0 +1,123 @@
...
+
+use function Flow\ETL\DSL\{array_to_rows};
+use Flow\ETL\Extractor\{FileExtractor, Limitable, LimitableExtractor, PathFiltering, Signal};
+use Flow\ETL\Row\Schema;
+use Flow\ETL\{Extractor, FlowContext};
+use Flow\Filesystem\{Path};
+use JsonMachine\Items;
+use JsonMachine\JsonDecoder\ExtJsonDecoder;
+
+final class JsonlExtractor implements Extractor, FileExtractor, LimitableExtractor
```
norberttech: I think I would call it JsonLinesExtractor, seems to be more self descriptive than just JsonL, wdyt?

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/functions.php
```diff
@@ -34,6 +34,19 @@ function from_json(
     return $loader;
 }
 
+/**
+ * @param Path|string $path - string is internally turned into stream
```
norberttech, line 40: can we add a link to https://jsonlines.org/ that explains the difference? The content of this docblock is turned later automatically into the `from_json_lines` documentation just like this: 

https://flow-php.com/documentation/dsl/json/from-json/#dsl-function

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/functions.php
```diff
@@ -34,6 +34,19 @@ function from_json(
...
+ */
+#[DocumentationDSL(module: Module::JSON, type: Type::EXTRACTOR)]
+#[DocumentationExample(topic: 'data_reading', example: 'json')]
+function from_json_lines(
+    string|Path $path,
+) : JsonlExtractor {
+    $loader = new JsonlExtractor(\is_string($path) ? Path::realpath($path) : $path);
+
+    return $loader;
```
norberttech: I would probably just do `return new JsonlExtractor(\is_string($path) ? Path::realpath($path) : $path);`

norberttech:
> Alright I've separated out the JSONLExtractor see what you think, once your happy ill do the loader separation.

I like it a lot! Left few minor comments but nothing critical, irrelevant details

src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JSONMachine/JsonlExtractor.php
jmortlock: definitely better

examples/topics/data_reading/jsonl/code.php
```diff
@@ -0,0 +1,24 @@
...
+    bool_schema('active'),
+);
+
+data_frame()
+    ->read(
+        from_json_lines(__DIR__ . '/input/dataset.json')
+            ->withSchema($schema)
+    )
+    ->collect()
+    ->write(to_stream(__DIR__ . '/output.txt', truncate: false))
```
norberttech, line 23: you might want to run that example locally and commit the output.txt as it would be rendered on the flow website

jmortlock:
I think I have addressed all the feedback, also extracted out the JsonLinesLoader.

Thanks for the prompt feedback is handy

norberttech:
Thank you @jmortlock looks awesome!! 🤩

Issue #1316, opened 2025-01-01
Integration with symfony/string for better strings manipulation


norberttech:
That would be very helpful integration from text processing point of view.

Library: https://github.com/symfony/string

It should be pretty straightforward since we would just need to create a scalar function for each of the methods string methods. 

I think we can make it a part of TextAdapter, just put all scalar functions inside of it.
~All those functions should be also added to TextAdapter DSL functions, with a text_ prefix.~
Those functions should exist in the `src/core/etl/src/Flow/ETL/Function` namespace and be part of the core ETL.

We don't even need to create for them dedicated DSL functions (as they can only operate on string), what we can do is simply expose them through `src/core/etl/src/Flow/ETL/Function/ScalarFunctionChain.php` so they would all be available through `ref()->xxxx` 

Documentation: [How to create scalar function](https://flow-php.com/blog/2024-08-08/scalar-functions/)

stloyd:
How does it differ from [Symfony String](https://symfony.com/doc/current/string.html#methods-added-by-codepointstring-and-unicodestring)?

norberttech:
Interesting, I would need to go through entire APIs of both libs to understand which one provides more value. 
symfony/string comes with some dependencies, that's
one thing.
However to provide more text manipulation options we should also take a look at https://github.com/voku/portable-utf8 

So the question is what makes more sense, portable-ascii + portable-utf8 or symfony/string?

norberttech:
I asked GitHub copilot which one we should use and he said symfony/string but without any good justification 😅 I wouldn't trust it, comparing the APIs - that's what we
need to do first

norberttech:
https://symfony.com/doc/current/string.html#methods-to-change-case just noticed that symfony/string can also replace an existing case converting dependency. 
So I guess we have our answer about which library to use

norberttech:
This is an example of how easy it is to create Scalar Functions: https://github.com/flow-php/flow/pull/1337

stloyd:
With recent PRs:
- #1635,
- #1617,
- #1607

I would say that this could be considered as done, no?

norberttech:
> I would say that this could be considered as done, no?

Hmm the initial idea was to cover all functions from [String Component](https://symfony.com/doc/current/string.html) by scalar functions, so I guess we are still missing few but I need to spend some time, go through list of functions and list what we already have and what we are still missing

PR #1449, opened 2025-02-05
feat: create Folded Scalar function with Test

f-lapinski:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>string folded function</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Ref: #1316


src/core/etl/src/Flow/ETL/DSL/functions.php
```diff
@@ -1688,3 +1689,9 @@ function caster(?Options $options = null) : Caster
 {
     return Caster::default($options ?? caster_options());
 }
+
+#[DocumentationDSL(module: Module::CORE, type: DSLType::SCALAR_FUNCTION)]
+function folded(ScalarFunction|string $value) : Folded
```
norberttech: I would not add this function to DSL. Those string functions they usually operate on a value, so the easiest use case for them is `ref('string_column')->stringFolded()`

norberttech: Actually I think the name should be `stringFold()`

src/core/etl/tests/Flow/ETL/Tests/Unit/Function/FoldedTest.php
```diff
@@ -0,0 +1,20 @@
...
+use function Flow\ETL\DSL\{folded, lit};
+use Flow\ETL\Tests\FlowTestCase;
+
+final class FoldedTest extends FlowTestCase
+{
+    public function test_string_folded() : void
+    {
+        self::assertSame(
+            "die o'brian strasse",
+            folded(lit("Die O'Brian Straße"))->eval(row())
```
norberttech: Ha, nice one! I would personally test it through a function chain, like this: 

```php
self::assertSame(
  "die o'brian strasse",
  ref("str")->stringFold()->eval(row(str_entry("str", "Die O'Brian Straße"))
)

```

norberttech:
@f-lapinski congratz on your first contribution 🥳

PR #1453, opened 2025-02-07
Add StringStyle ScalarFunction

f-lapinski:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    

<li>Change StringCamel to StringStyle to using with exisitng Enum StringStyles and get's StringStyle as parameter</li>

  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
<li>Previous StringCamle function, methods and tests</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Ref: #1316

<p>The StringStyle function has been added to the codebase with an implemented test to ensure functionality. It has been integrated into the ScalarFunctionChain, enabling its use within the chain of scalar functions. Working with StringStyle enum's.</p>


f-lapinski:
`1) Flow\ETL\Tests\Unit\Function\StringStyleTest::test_string_style_kebab
Error: Call to undefined method Symfony\Component\String\UnicodeString::kebab()`

Test throwing error about string method kebab(), cause it's avaliable on symfony 7.2+
There is two options i think, move Smyofny-string package to 7.2 or exclude kebab from tests.

src/core/etl/src/Flow/ETL/Function/StringStyle.php
```diff
@@ -0,0 +1,34 @@
...
+
+use function Flow\ETL\DSL\type_string;
+use Flow\ETL\Function\ScalarFunction\TypedScalarFunction;
+use Flow\ETL\Function\StyleConverter\StringStyles;
+use Flow\ETL\PHP\Type\Type;
+use Flow\ETL\Row;
+
+final class StringStyle extends ScalarFunctionChain implements TypedScalarFunction
+{
+    public function __construct(private readonly ScalarFunction|string $string, private readonly StringStyles $style)
```
norberttech: this should be a `ScalarFunction|string|StringStyles` so you can use it like this: 

ref()->stringStyle('snake')
ref()->stringStyle(StringStyles::SNAKE)
ref()->stringStyle(ref('style'))

norberttech:
> `1) Flow\ETL\Tests\Unit\Function\StringStyleTest::test_string_style_kebab Error: Call to undefined method Symfony\Component\String\UnicodeString::kebab()`
> 
> Test throwing error about string method kebab(), cause it's avaliable on symfony 7.2+ There is two options i think, move Smyofny-string package to 7.2 or exclude kebab from tests.

So if you look at `u()->kebab()` this is what you are going to see: 

```
   public function kebab(): static
    {
        return $this->snake()->replace('_', '-');
    }
```

So all we need to do is at the level of StringStyles, check if `kebab` method is available on UnicodeString and if not do `$this->snake()->replace('_', '-')`

src/core/etl/src/Flow/ETL/Function/StringStyle.php
f-lapinski: done :)

src/core/etl/src/Flow/ETL/Function/StringStyle.php
```diff
@@ -0,0 +1,42 @@
...
+    ) {
+    }
+
+    public function eval(Row $row) : mixed
+    {
+
+        $string = (new Parameter($this->string))->asString($row);
+        $style = (new Parameter($this->style))->as($row, type_string(), type_enum(StringStyles::class));
+
+        if ($string === null) {
```
norberttech: you also need to check if style is not null here

PR #1469, opened 2025-02-11
Don't track write count in the JSONLinesLoader

jmortlock:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>JSONLines Loader would occasionally write a newline to the start of the file.</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

In my local usage of this I would occasionally get a newline at the start of the file which causes upstream systems to reject the file, now I could not reproduce this in the test case and during my investigation of this I found I can simplify the writing  by not tracking the write count, which solves the newline at start of file issue and simplifies the code.

In JSONL specification
```
The last character in a file following the last JSON value may be a line separator. In this case the line separator does not indicate the start of another JSON value. 
```
This means we can happily writeln without worry that the last line is just \n

### DIFF
--- a/src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JsonLinesLoader.php
+++ b/src/adapter/etl-adapter-json/src/Flow/ETL/Adapter/JSON/JsonLinesLoader.php
@@ -15,11 +15,6 @@ final class JsonLinesLoader implements Closure, Loader, Loader\FileLoader
 
     private int $flags = JSON_THROW_ON_ERROR;
 
-    /**
-     * @var array<string, int>
-     */
-    private array $writes = [];
-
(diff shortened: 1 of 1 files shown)


norberttech:
hey @jmortlock sorry for super late response, was out most of the day, look great! 🙌

Issue #1472, opened 2025-02-12
Doctrine Insert with XmlEntry


jmortlock:
Hi

If i try and insert a dataframe into a database that contains an XmlEntry, I get the following error:

```
Error: Object of class DOMDocument could not be converted to string

/var/www/html/zg/vendor/doctrine/dbal/src/Driver/PDO/Statement.php:48
/var/www/html/zg/vendor/doctrine/dbal/src/Connection.php:1809
/var/www/html/zg/vendor/doctrine/dbal/src/Connection.php:1205
/var/www/html/zg/vendor/doctrine/dbal/src/Connections/PrimaryReadReplicaConnection.php:292
/var/www/html/zg/vendor/flow-php/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/Bulk.php:42
/var/www/html/zg/vendor/flow-php/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/DbalLoader.php:67
/var/www/html/zg/vendor/flow-php/etl/src/Flow/ETL/Pipeline/SynchronousPipeline.php:65
/var/www/html/zg/vendor/flow-php/etl/src/Flow/ETL/DataFrame.php:763
```

Some pseduo code

```sample.json
{"descriptionHtml": "<b>This is a test</b>"}
```

```
       return data_frame()
            ->read(from_json_lines("sample.json")
            ->write(to_dbal_table_insert(.....)
```

I had a quick look at submitting a fix, as I managed to hack the v0.10.0 code to get it to work, however looks like I need a postgress db to run the test-suite which i don't currently have installed, and i suspect this is already fixed or easy enough for you.

I believe you need to handle it here

https://github.com/flow-php/flow/blob/dfde1cc8238a42281f77a60f44851052ca34c885/src/lib/doctrine-dbal-bulk/src/Flow/Doctrine/Bulk/BulkData.php#L137

I believe something like this
```
\DOMDocument::class => $entry->saveXML($entry->documentElement),
```

As i side note are they are workarounds you can think of to transform the datatype in the row?

Closest I got was
```
->withEntry('DESCRIPTION_HTML', ref('DESCRIPTION_HTML')->domElementValue())
```
But that strips out the html tags, which ideally i want to keep.

norberttech:
Hey!
What you are doing with ref()->domElementValue() is the correct approach.
I'm not sure why it's stripping HTML tags, as this function just takes nodeValue. 
[...]

PR #1474, opened 2025-02-12
Failing example for Xml insert

jmortlock:
Example of trying to insert an XmlEntry into doctrine.

See #1472

### DIFF
--- a/src/adapter/etl-adapter-doctrine/tests/Flow/ETL/Adapter/Doctrine/Tests/Integration/DbalLoaderTest.php
+++ b/src/adapter/etl-adapter-doctrine/tests/Flow/ETL/Adapter/Doctrine/Tests/Integration/DbalLoaderTest.php
@@ -6,7 +6,7 @@ namespace Flow\ETL\Adapter\Doctrine\Tests\Integration;
 
 use function Flow\ETL\Adapter\Doctrine\{to_dbal_table_insert, to_dbal_table_update};
 use function Flow\ETL\DSL\data_frame;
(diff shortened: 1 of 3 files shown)


norberttech:
thank you @jmortlock I'm gonna take a look a bit later today!

norberttech:
hey @jmortlock I played with your tests, and the reason why they are failing is because you are using `DOMDocument::loadHTML()` which adds HTML tags in order to make it a valid html document. 
If you use `DOMDocument::loadXML()` the html part shouldn't be there anymore

jmortlock:
I've updated the tests to use loadXml there all still failing.

I guess the question that comes to my mind, if we are using JSON for example do we actually want to auto infer DOMDocument / Xml ? I know for my use cases which is embded html in a json document I really just expect a string, if i wanted to purify/sanitize/strip_tags those apis are also expecting string inputs.

norberttech:
> I guess the question that comes to my mind, if we are using JSON for example do we actually want to auto infer DOMDocument / Xml ? I know for my use cases which is embded html in a json document I really just expect a string, if i wanted to purify/sanitize/strip_tags those apis are also expecting string inputs.

I would keep this behavior as it's a generic one, unrelated to source/destination (it comes from EntryFactory) and in your case `->withSchema()` and tell the EntryFactory that given column is just a string

norberttech:
@jmortlock, but based on your work, I was able to find inconsistent behavior in casting XMLs to strings. I'm closing this PR and will open another one with fixes (that will include your commits and keep your contribution).

PR #1477, opened 2025-02-13
Add Title function with allWords parameter set default to false.

f-lapinski:
Add Test for function with and without allWords and return type

<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
<li>Add Title function with argument allWords set false as default with Tests</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Ref: #1316

<p>Function title() and addWords parameter set false on default and add Test for title function with and without allWords and test for return type</p>


norberttech:
Good job 👏

PR #1481, opened 2025-02-14
Add localTitle function with Test for 'en' and 'nl'

f-lapinski:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
<li>localeTitle function added with test coverage of 'en' and 'nl' examples and return type
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

<p>Added function localeTitle() changes all graphemes/code points to "title case" according to locale-specific case mappings.

For example:
<li>- localeTitle('en')</li>
</p>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Function/ScalarFunctionChain.php
+++ b/src/core/etl/src/Flow/ETL/Function/ScalarFunctionChain.php
@@ -487,6 +487,11 @@ abstract class ScalarFunctionChain implements ScalarFunction
         return new StringFold($this);
     }
 
+    public function stringLocaleTitle(ScalarFunction|string $locale) : self
+    {
+        return new StringLocaleTitle($this, $locale);
+    }
+
     public function stringStyle(ScalarFunction|string|StringStyles $style) : self
     {
         return new StringStyle($this, $style);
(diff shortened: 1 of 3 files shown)


norberttech:
hey! Let me think about this one as I'm not fully sure how to handle intl dependency in this case, need to check if we are not bringing a polyfill already anyway.

f-lapinski:
Sure, no problem :)

stloyd:
As additional comment I have already had that issue with adding "yet more option/required" extension, and my thought was that maybe it's time to make new package like "ETL-extra"/"etl-string-functions" or whatever would be best fit, but to move most "depending" functions out of main package, same way we have adapters or bridges.

norberttech:
Yeah, moving functions based on extensions/libraries to standalone packages (libs/bridges) is the way to go.
My only challenge here is how to make it part of the ScalaFunctionChain so IDE can autocomplete it when we type ref('...')-> 🤔

norberttech:
I'm closing this one for now, I'm not sure yet how to tackle this but it's also not the highest priority

PR #1482, opened 2025-02-15
Fix double encoding of json

jmortlock:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>JSON was being double encoded.</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

I was looking at replacing ArrayEntry with JsonEntry and hit this issue with the JSON serialisation in which it would double encode the JSON result (once in the normaliser, than once more in the loader)


stloyd:
I would say that this is tricky, cause you may want to double encode json inside json in some edge cases as well.

jmortlock:
Then ArrayEntry still has a place?

What are the scenarios where we need to double encode? You could just use a string type in that case?

norberttech:
> Then ArrayEntry still has a place?

No, array entry was entirely removed and it's not coming back, JsonEntry holds an array under the hood.

I don't see any good reason to keep double encoding, this behavior is a leftover, but I might be missing something 

@stloyd you have any specific scenarios in your head?

norberttech:
I merged it for now as I don't yet see how double encoding can be useful, but feel free to ping me if you can think about anything, we can still revert that change or parametrize it

PR #1541, opened 2025-03-20
Allow to duplicate row

### LINKED ISSUE #1539: Duplicate Row
Sometimes we need to simply duplicate a row based on specific condition and then apply some transformations on it (or not). 

Currently the only way to achieve that would be through iterating through dataset, filtering the rows we want to modify, apply transformations and then do union merge it with the main data frame. 

But there is a better way:

```php
<?php
$result = DataFrame::df()
    ->read(from_array($data))
    ->duplicateRow(
        condition: ref('status')->equals('canceled'),
        transformations: [
           with_entry('amount', ref('amount')->multiply(-1)),
        ]
    )
    ->run()
    ->toArray();
```

We would need to introduce new dsl function, `with_entry(string|Reference $ref, ScalarFunction $transformation)`

That scalar function would simply create WithEntry object that we can also pass into DataFrame::transform() method alongside Transformer|Transformation|Transformations 
which would be another great way to chain many transformations. 

It would also allow us to deprecate `array<string, ScalarFunction|WindowFunction>` in withEntries and replace it with `array<WithEntries>`

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>DataFrame::duplicateRow() method</li>
    <li>with_entry() function</li> 
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>DataFrame::with/transform/withEntries now also accepts WithEntry instance</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Resolves #1539


src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/TransformTest.php
```diff
@@ -0,0 +1,135 @@
...
+            ->toArray();
+
+        self::assertEqualsWithDelta(
+            [
+                ['id' => 1, 'status' => 'active', 'amount' => 410.00, 'currency' => 'PLN'],
+                ['id' => 2, 'status' => 'inactive', 'amount' => 2.5, 'currency' => 'PLN'],
+                ['id' => 3, 'status' => 'active', 'amount' => 10.00, 'currency' => 'PLN'],
+            ],
+            $rows,
+            0.01
```
bendavies: why is this needed?

norberttech: great catch! 
It's not, I think I copied it from another place that was actually failing due to floating point comparison

PR #1577, opened 2025-04-05
Fixed merging float definitions with different precisions

### LINKED ISSUE #1576: Handling Floats
Currently flow math functions are operating on float's directly which as we all know in php can be a bit tricky. 

For example 0.3 - 0.2 is not the same as 0.1.

However flow is partially covering that since it operates on entries and FloatEntry comes with a precision. 

There are however few scenarios where this wont be enough: 

- merging float_schema definitions with different precisions 
- math operations on float entries with precision higher than 6 
- using math scalar functions on entries with precision higher than 6

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
    <li>ScalarFunctionResult</li>
    <li>bcmath extension with polyfill</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Merging float definitions with different precisions</li>
    <li>Merging floats and integer to use float precision</li>
    <li>Merging lists/maps of floats to use higher precision</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Float type toString now display precision</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>TypeScalarFunctionInterface</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

In general all ScalarFunctions can return any value `mixed`. This means that the result of scalar function needs to go through `EntryFactory` in order to create an entry. 

Scalar functions can take their values from:

- reference - value is going to be taken from entry (we do have access to type)
- literal - we don't have access to type, just value 
- scalar function - the value is a result of evaluation of another scalar function - we dont have access to type 

This means that functions like plus/minus/divide/multiply/mod will always return some float|integer and they wont keep the precision/scale of the input float (if the input was an entry). 

In order to fix that, scalar functions can now also return `ScalarResult(mixed $value, Type $type)` which is pretty much just a wrapper around the value. 

This approach is way more flexible that TypedScalarFunction since some return types are dynamic and depends on input arguments, for example `ref('id')->cast(type_int())`. 

The return type of cast depends on the argument. 

So now when ScalarResult is returned function can pass over more data, like for example type taken from the input. 

But since not all values must come from a entry, all math functions received another argument `ScalarFunction|int $scale = 0` which can be used to specify how precise that operation needs to be. Under the hood, scale is passed to the instance of `Calculator`.


src/core/etl/src/Flow/ETL/Function/Divide.php
```diff
@@ -4,29 +4,52 @@
...
+
+        $scale = max($leftScale, $rightScale);
+
+        // we know that the result is going to be float so if the scale wasn't set
+        // intentionally we set it to 6
+        if ($scale === 0 && $this->scale === null && ($leftValue % $rightValue) !== 0) {
+            $scale = 6;
+        }
+
+        var_dump(\json_encode(['scale' => $scale, 'left_scale' => $leftScale, 'right_scale' => $rightScale, 'left_value' => $leftValue,  'right_value' => $rightValue]));
```
bendavies, line 46: debug left in

bendavies: (can be automated with https://github.com/spaze/phpstan-disallowed-calls)

norberttech: haha it happens to me way too often! Going to add that extension now 🤣

Thats why I release on a schedule, to give myself time to fix those kind of mistakes 🙈

PR #1606, opened 2025-04-28
[GoogleSheet] Prevent fatal error when extracted columns don't match headers amount

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>[GoogleSheet] Prevent fatal error when extracted columns don't match headers amount</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

In edge cases when the given spreadsheet has some random columns filled (i.e., as comments from the reporter), it will fail to match properly the headers with columns, leading to a fatal error:
```
array_combine(): Argument #1 ($keys) and argument #2 ($values) must have the same number of elements
```


src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/functions.php
```diff
@@ -25,7 +24,7 @@ function from_google_sheet(
     bool $with_header = true,
     int $rows_per_page = 1000,
     array $options = [],
-) : Extractor {
+) : GoogleSheetExtractor {
```
stloyd, line 27: This class cannot be different, and this prevents end-user phpstan errors like:
```
Potentially polymorphic call. The code may be inoperable depending on the actual class instance passed as the argument. 
```
When running code like:
```php
from_google_sheet(
                        json_decode(file_get_contents(__DIR__.'/file.json), true),
                        $spreadsheetId,
                        'sheet-name',
                    )->withDropExtraColumns(true)

File: src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php [hunk 2]
```diff
@@ -57,24 +59,30 @@ public function extract(FlowContext $context) : \Generator
             $totalRows = 0;
         }
 
+        $headersCount = \count($headers);
```
stloyd, line 62: Small boost when you have a looooooot of rows...

src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration/GoogleSheetExtractorTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Row has more columns (4) than headers (3)');
+
+        foreach ($rows as $row) {
+            self::assertNotNull($row);
+        }
+    }
+
+    private function createHttpClient() : Client
```
norberttech: I would move this to GoogleSheetsContext that extends from FlowContext and let it take path to fixture file or alternatively also associative array that would be change to json on the fly (union type arg maybe).
This would make it easier to cover other things with tests

norberttech: and I would make it return Google Sheet client hiding the http client existence entirely

norberttech (changes requested):

norberttech (approved):

PR #1608, opened 2025-04-28
Allow custom way of renaming values

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Allow custom way of renaming values</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

As provided in test examples, we can easily apply custom renaming, i.e., translitering non-Latin characters to Latin ones with a callback function.

### DIFF
--- a/src/core/etl/src/Flow/ETL/DataFrame.php
+++ b/src/core/etl/src/Flow/ETL/DataFrame.php
@@ -637,6 +637,16 @@ final class DataFrame
     /**
      * @lazy
      */
+    public function renameAllCallback(\Closure $callback) : self
+    {
+        $this->pipeline->add(new RenameAllCaseTransformer(callback: $callback));
+
+        return $this;
+    }
+
+    /**
+     * @lazy
+     */
     public function renameAllLowerCase() : self
     {
         $this->pipeline->add(new RenameAllCaseTransformer(lower: true));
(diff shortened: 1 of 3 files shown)


src/core/etl/src/Flow/ETL/DataFrame.php
```diff
@@ -634,6 +634,16 @@ public function renameAll(string $search, string $replace) : self
         return $this;
     }
 
+    /**
+     * @lazy
+     */
+    public function renameAllCallback(\Closure $callback) : self
```
norberttech, line 640: I'm not a fan of adding more callbacks to data frame. I started working on a parallel processing on a side and each api method that adds callback is just a pure hell since in parallel processing I need to serialize and spread "transformations" among many processes which with callbacks is extremely unstable.

Even for map I had to put an exception that DataFrame::map() cant be used with parallel processing and that its recommended to implement a scalar function or transformation instead.

PR #1609, opened 2025-04-28
[GoogleSheet] Add a new Google Sheet loader

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>[GoogleSheet] Add a new Google Sheet loader</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/RowsNormalizer/EntryNormalizer.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function normalize(Entry $entry) : string|float|int|bool|null
+    {
+        return match ($entry::class) {
+            Entry\UuidEntry::class,
+            Entry\XMLElementEntry::class,
+            Entry\XMLEntry::class => $entry->toString(),
+            Entry\DateTimeEntry::class => $entry->value()?->format($this->dateTimeFormat),
+            Entry\DateEntry::class => $entry->value()?->format($this->dateFormat),
+            Entry\TimeEntry::class => $entry->value() ? date_interval_to_microseconds($entry->value()) : null,
```
norberttech: I believe excel has a decent format for intervals that can be rendered like `00:00`

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/RowsNormalizer/EntryNormalizer.php
```diff
@@ -0,0 +1,40 @@
...
+
+use function Flow\ETL\DSL\{date_interval_to_microseconds, type_json};
+use Flow\ETL\PHP\Type\Caster;
+use Flow\ETL\Row\Entry;
+
+final readonly class EntryNormalizer
+{
+    public function __construct(
+        private Caster $caster,
+        private string $dateTimeFormat = \DateTimeInterface::ATOM,
```
norberttech: Is this the format that googl sheets recognize automatically as timestamp?

stloyd: This can be handled if we allow passing as user entered for input value in update, if we go with raw we should manipulate the data.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,100 @@
...
+
+    public function load(Rows $rows, FlowContext $context) : void
+    {
+        if (!$rows->count()) {
+            return;
+        }
+
+        $sheetRange = SheetRangeCalculator::calculate(
+            count($rows->entries()),
+            $rows->count() + ($this->withHeader ? 1 : 0),
```
norberttech: that might not be enough, for Rows with only one Row it will keep overwriting it since you are not moving the starting position beyond already written rows

src/adapter/etl-adapter-google-sheet/composer.json
```diff
@@ -28,7 +28,7 @@
             ]
         },
         "files": [
-            "src/Flow/ETL/Adapter/GoogleSheet/functions.php"
+            "src/Flow/ETL/Adapter/GoogleSheet/DSL/functions.php"
```
norberttech, line 31: this requires a note in documentation/upgrading.md

stloyd: So maybe revert?

src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration/GoogleSheetTest.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    protected function tearDown() : void
+    {
+        parent::tearDown();
+
+        $this->clearTestSheet();
+    }
+
+    public function test_load_and_extract() : void
```
stloyd, line 62:
```bash
/usr/local/Cellar/php@8.2/8.2.28_1/bin/php/Users/stloyd/Documents/flow/tools/phpunit/vendor/phpunit/phpunit/phpunit --configuration /Users/stloyd/Documents/flow/phpunit.xml.dist --filter Flow\\ETL\\Adapter\\GoogleSheet\\Tests\\Integration\\GoogleSheetTest --test-suffix GoogleSheetTest.php /Users/stloyd/Documents/flow/src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration --teamcity
Testing started at 11:23 ...
PHPUnit 11.5.18 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.28
Configuration: /Users/stloyd/Documents/flow/phpunit.xml.dist
Time: 00:05.952, Memory: 16.00 MB

OK (1 test, 7 assertions)
Process finished with exit code 0

File: src/adapter/etl-adapter-google-sheet/composer.json (hunk 4 above)
Comment (norberttech) [reply src/adapter/etl-adapter-google-sheet/composer.json:31]: nah, it's better this way and this adapter is not yet popular enough so it's probably best time for such change, just put one line in upgrading.md and will be perfect

File: src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration/GoogleSheetTest.php [hunk 6]
```diff
@@ -0,0 +1,98 @@
...
+        );
+
+        $loader->load(
+            rows(
+                row(int_entry('id', 12345), string_entry('name', 'Norbert')),
+                row(int_entry('id', 54321), string_entry('name', 'Joseph')),
+                row(int_entry('id', 666), string_entry('name', 'Dominik'))
+            ),
+            flow_context(config())
+        );
```
norberttech, line 84: I would change it to: 

```php
        $loader->load(
            rows(
                row(int_entry('id', 12345), string_entry('name', 'Norbert')),
                row(int_entry('id', 54321), string_entry('name', 'Joseph')),
            ),
            $context = flow_context(config())
        );
        $loader->load(
            rows(
                row(int_entry('id', 666), string_entry('name', 'Dominik'))
            ),
            $context
        );        
```

To make sure that it wont overwrite previous rows

src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration/GoogleSheetTest.php
norberttech: it's with auth env var set?

src/adapter/etl-adapter-google-sheet/composer.json
stloyd: Added.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,136 @@
...
+use Google\Service\Sheets;
+use Google\Service\Sheets\ValueRange;
+
+final class GoogleSheetLoader implements Loader, Loader\Closure
+{
+    private ?int $columnsCount = null;
+
+    private string $dateFormat = 'Y-m-d';
+
+    private string $dateTimeFormat = \DateTimeInterface::ATOM;
```
norberttech: not sure about google sheets but excel prefer `Y-m-d H:i:s` over ATOM (`Y-m-d\TH:i:sP`) and it automatically detects it as a date time

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,136 @@
...
+
+    private int $startingRow = 0;
```
norberttech: `$startingRow` -> `$loadedRows`

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,136 @@
...
+        $this->service->spreadsheets_values->update(
+            $this->spreadsheetId,
+            $sheetRange->toString(),
+            new ValueRange(
+                [
+                    'values' => $this->values($rows, $normalizer),
+                    'majorDimension' => 'ROWS',
+                ]
+            ),
+            ['valueInputOption' => 'RAW']
```
norberttech: I would let end user to parametrize it, it's pretty convinient in many cases to let excel/google do the auto formatting: 

https://developers.google.com/workspace/sheets/api/reference/rest/v4/ValueInputOption?hl=pl

We just need an enum for that, you can literally copy description of each field to that enum to make it easier to understand

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,136 @@
...
+
+        return $this;
+    }
+
+    private function calculateCurrentRow(Rows $rows) : int
+    {
+        return $rows->count() + (0 === $this->startingRow ? ($this->withHeader ? 1 : 0) : $this->startingRow);
+    }
+
+    private function calculateSheetRange(int $currentRow) : SheetRangeCalculator
```
norberttech: hmm the method says `calculte` but it returns `SheetRangeCalculator`, I would rename `SheetRangeCalculator` into `SheetRange`

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
stloyd: During debug I removed the param, enum is there already 😂

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public function withHeader(bool $withHeader) : self
+    {
+        $this->withHeader = $withHeader;
+
+        return $this;
+    }
+
+    public function withStartingRow(int $startingRow) : self
```
norberttech: so we can let user to decide to start writing from row 100, will it also put properly a header when custom starting row is set? 

Also if we want to allow users to set different starting row we should also let them set column different than A

norberttech: Actually, maybe we just let users to pass `SheetRange`?

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
stloyd: We have `SheetRange` class already, but it's working a bit differently, so that's why I added a new one, as the scope of this PR is getting really wide :)

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
stloyd: No, it was for internal testing purpose; it should not be exposed.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
norberttech: >scope of this PR is getting really wide :) 

Maybe, but I already made that mistake many times in the past that I merged half-baked solutions and at the end I was the one who needed to maintain/improve them over time. So as much as inconvenient and unnecessary it might look, I would rather prefer to merge only complete and usable solutions

norberttech: > We have SheetRange class already, 

Uhm, not great, we should then probably merge them or use another name like `SheetAddress`

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
norberttech: at some point it might be a nice addition to let end users to change the starting address but it can come later once someone will get first use case for it

norberttech:
One last thing we will need in this PR is an example in examples, to your point that we are missing a lot of docs/examples 😅

src/adapter/etl-adapter-google-sheet/tests/Flow/ETL/Adapter/GoogleSheet/Tests/Integration/GoogleSheetTest.php
stloyd: Replaced with a file-based one as we spoke before.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
stloyd: Renamed.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,130 @@
...
+
+    private string $timeFormat = 'H:i';
+
+    private bool $withHeader = true;
+
+    public function __construct(
+        private readonly Sheets $service,
+        private readonly string $spreadsheetId,
+        private readonly string $sheetName,
+        private readonly ValueInputOption $inputOption,
```
norberttech: This should be passed through `->withInputOption` not through constructor

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,130 @@
...
+    private int $loadedRows = 0;
+
+    private string $timeFormat = 'H:i';
+
+    private bool $withHeader = true;
+
+    public function __construct(
+        private readonly Sheets $service,
+        private readonly string $spreadsheetId,
+        private readonly string $sheetName,
```
norberttech, line 34: why do we need sheet name when there is spreadsheetid?

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/RowsNormalizer/EntryNormalizer.php
```diff
@@ -0,0 +1,41 @@
...
+        private string $dateFormat = 'Y-m-d',
+        private string $timeFormat = 'H:i',
```
norberttech: those default values are no longer needed, loader is passing them  anyway

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
stloyd: Sheet name can be different, i.e., you can have two sheets in one spreadsheet:
Zrzut ekranu 2025-05-3 o 14 05 54
You need to point which one you want to update.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,138 @@
...
+
+    private function calculateCurrentRow(Rows $rows) : int
+    {
+        return $rows->count() + (0 === $this->loadedRows ? ($this->withHeader ? 1 : 0) : $this->loadedRows);
+    }
+
+    private function calculateSheetAddress(int $currentRow) : SheetAddress
+    {
+        return SheetAddress::calculate(
+            (int) $this->columnsCount,
```
norberttech: Shouldn't this be taken from the Rows that are currently being loaded? 
Currently this value comes from the first row, but there is no guarantee that another row will not add more columns.

examples/topics/data_writing/google_sheet/.gitignore
```diff
@@ -0,0 +1,2 @@
+auth.json.dist
```
norberttech: it should be auth.json that is ignored not .dist - in fact .dist should be committed

stloyd: I needed to limit it slightly as GH blocked the commit due to security checks.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,179 @@
...
+        $this->service->spreadsheets->batchUpdate(
+            $this->spreadsheetId,
+            new BatchUpdateSpreadsheetRequest([
+                'requests' => [
+                    [
+                        'updateSheetProperties' => [
+                            'properties' => [
+                                'sheetId' => $this->sheetId(),
+                                'gridProperties' => [
+                                    'rowCount' => $currentRow + 5_000,
```
norberttech, line 130: shouldn't this just add count of loaded Rows? Otherwise it will just keep allocating 5k rows even when it's not necessary

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,179 @@
...
+                        ],
+                    ],
+                ],
+            ])
+        );
+    }
+
+    private function isDividable(int $currentRow) : bool
+    {
+        return 0 === $currentRow % 5000;
```
norberttech: seriously? 😂 I get that you want to merge it ASAP but lets keep some quality 😅

This method is not only not doing what it says it's doing but also it keeps the default 5k hardcoded when in reality we might be updating a sheet that someone manually reduced to 1k rows

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,179 @@
...
+
+        $spreadsheet = $this->service->spreadsheets->get($this->spreadsheetId);
+
+        foreach ($spreadsheet->getSheets() as $sheet) {
+            if ($sheet->getProperties()->getTitle() === $this->sheetName) {
+                return $this->sheetId = $sheet->getProperties()->getSheetId();
+            }
+        }
+
+        throw new \RuntimeException("Sheet '{$this->sheetName}' not found");
```
norberttech: it should be exception from Flow namespace

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetLoader.php
```diff
@@ -0,0 +1,179 @@
...
+                                ],
+                            ],
+                            'fields' => 'gridProperties(rowCount,columnCount)',
+                        ],
+                    ],
+                ],
+            ])
+        );
+    }
+
```
norberttech: I'm not suer if that's the correct logic here, here is how I would do the dynamic rows allocation. 

1) when load is called for the first time get the current size of spreadsheet 
2) before loading rows check if we are not going to exceed the size of spreadsheet
  a) if spreadsheet is not large enough - expand it but 1k - but allow to modify this through `withSheetDynamicAllocationSize()` 
  b) if there is enough space just keep loading 
  
Additionally with each load you should also check if number of entires is smaller than number of allocated columns, if it's not, you should also add columns.

PR #1615, opened 2025-05-03
Add a new `DataFrame::renameEach()`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `DataFrame::renameEach()`</li>
    <li>Add a new functions `rename_style()`, `rename_replace()` & `rename_transliterate()`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    <li>Deprecate `Flow\ETL\DataFrame::renameAll()`</li>
    <li>Deprecate `Flow\ETL\DataFrame::renameAllLowerCase()`</li>
    <li>Deprecate `Flow\ETL\DataFrame::renameAllUpperCase()`</li>
    <li>Deprecate `Flow\ETL\DataFrame::renameAllUpperCaseFirst()`</li>
    <li>Deprecate `Flow\ETL\DataFrame::renameAllUpperCaseWord()`</li>
    <li>Deprecate `Flow\ETL\RenameAllCaseTransformer`</li>
    <li>Deprecate `Flow\ETL\RenameStrReplaceAllEntriesTransformer`</li>
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

New way of renaming all entries:
```php
df()->renameEach(rename_transliterate("..."));
df()->renameEach(rename_style(Style::LOWER));
df()->renameEach(rename_style(Style::UCFIRST));

### CODE REVIEW

File: src/core/etl/src/Flow/ETL/Transformer/RenameAllTransliterateTransformer.php [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+
+use Flow\ETL\{FlowContext, Row, Rows, Transformer};
+
+final readonly class RenameAllTransliterateTransformer implements Transformer
+{
+    public function transform(Rows $rows, FlowContext $context) : Rows
+    {
+        return $rows->map(function (Row $row) : Row {
+            foreach ($row->entries()->all() as $entry) {
+                $row = $row->rename($entry->name(), (string) \transliterator_transliterate('Any-Latin; Latin-ASCII; Lower()', $entry->name()));
```
norberttech: `transliterator_transliterate` doesn't that require an intl extension?

norberttech: `'Any-Latin; Latin-ASCII; Lower()'` should be probably moved to constructor parameter

src/core/etl/src/Flow/ETL/DataFrame.php
```diff
@@ -656,6 +657,16 @@ public function renameAllStyle(StringStyles|string $style) : self
         return $this;
     }
 
+    /**
+     * @lazy
+     */
+    public function renameAllTransliterate() : self
```
norberttech: uhm I think I made a mistake adding all of those `rename` methods to DataFrame API. 
It's not only breaking open close principle but also will keep growing and use cases like this one (that depends on php extensions) are going to throw exceptions when methods are not going to be available. 

I think we should change that approach and keep only two methods: 

```
DataFrame::rename(string $from, string $to) : self
DataFrame::renameEach(RenameStrategy $strategy) : self 
```

This way we can replicate all of the currently existing methods as a strategies. 

So there are 2 options here: 

1) you wait for me to rework the renaming API and then you move this transformer under new API. (I should be able to do it later next weekend or the week after next weekend) 

2) you create that new API and move the transformer under it

I would like to avoid merging this, adding new method to the API in next release and removing it in the following one. 

Thoughts?

stloyd: Sounds similar to something I had in mind when I looked at all those methods. Reworked to apply the first rename to each.

src/core/etl/src/Flow/ETL/Transformer/StyleConverter/RenameStrategy.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Transformer\StyleConverter;
+
+enum RenameStrategy : string
```
norberttech: that's not exactly what I had in mind, since it does not allow for any paramters and transliterate would need one. 
Also it moves problem just to another place, instead of having methods now we have enum cases. 

I was thinking about making it more sophisticated where `RenameStrategy` is an interface and each of the below can be it's own implementation. 

This way it can be used like this: 

```
df()->renameEach(rename_transliterate("..."))
df()->renameEach(rename_style(Style::LOWER_CASE))
df()->renameEach(rename_style(Style::UCFIRST))
```

src/core/etl/src/Flow/ETL/Transformer/StyleConverter/Style.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Transformer\StyleConverter;
```
norberttech: I would change that namespace to `Flow\ETL\Transformer\Rename` as this is not converting style only anymore

src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/RenameTest.php
```diff
@@ -98,6 +103,24 @@ public function test_rename_all_to_snake_case() : void
...
+
+        self::assertEquals(
+            [
+                ['osmy' => 8],
+                ['dziewiaty' => 9],
+            ],
+            \iterator_to_array($ds)
+        );
+    }
+
```
norberttech, line 122: dont you also need to replace those deprecated methods in remaining tests with the renemaeEach to avoid static analysis complaining about usage of deprecated methods?

src/core/etl/src/Flow/ETL/Transformer/RenameEachTransformer.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Flow\ETL\Transformer;
+
+use Flow\ETL\{FlowContext,
+    Row,
+    Rows,
+    Transformer,
+    Transformer\StyleConverter\RenameEntryStrategy
+};
+
+final readonly class RenameEachTransformer implements Transformer
```
norberttech: To keep it consistent with RenameEntryTransfromer this one should be called RenameEachEntryTransformer

documentation/upgrading.md
```diff
@@ -5,13 +5,34 @@ Please follow the instructions for your specific version to ensure a smooth upgr
...
+- `Flow\ETL\DataFrame::renameAllUpperCaseFirst()`,
+- `Flow\ETL\DataFrame::renameAllUpperCaseWord()`,
+
+Were deprecated in favor of using new method: `DataFrame::renameEach()` with proper `RenameEntryStrategy` object.
+
+### 2) Deprecated `RenameAllCaseTransformer` & `RenameStrReplaceAllEntriesTransformer`
+
+Selected transformers were deprecated in favor of using `DataFrame::renameEach()` with related `RenameEntryStrategy`:
+- `RenameAllCaseTransformer` -> `RenameCaseTransformer`,
+- `RenameStrReplaceAllEntriesTransformer` -> `RenameReplaceStrategy`,
```
norberttech, line 25: I'm adding comment to create task from it to remove those deprecated classes before 1.0

norberttech (approved):

src/core/etl/tests/Flow/ETL/Tests/Integration/DataFrame/RenameTest.php
stloyd: Split test + added phpunit specific attribute on deprecated tests.

PR #1635, opened 2025-05-07
Merge `Rename\Style` with `StringStyles`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>`StringStyles` related code to work with UTF-8</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Merge `Rename\Style` with `StringStyles`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    <li>`Flow\ETL\Function\StyleConverter\StringStyles` in favor of `Flow\ETL\Style\StringStyles`</li>
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/core/etl/src/Flow/ETL/DSL/functions.php
```diff
@@ -349,7 +348,7 @@ function to_branch(ScalarFunction $condition, Loader $loader) : Loader\Branching
 }
 
 #[DocumentationDSL(module: Module::CORE, type: DSLType::TRANSFORMER)]
-function rename_style(Style $style) : Transformer\Rename\RenameCaseEntryStrategy
+function rename_style(StringStyles $style) : Transformer\Rename\RenameCaseEntryStrategy
```
norberttech: It's a hard BC break, it also creates a coupling between Scalar Function and Rename Transformer. 

I would mve that Stringstyles class to `Flow\ETL\String\StringStyles` namespace first keeping those two others and making them deprecated. 
The I would do a union type here and in that function `Style|StringStyles` and mark old classes as deprecated which I will remove before 1.0 release

stloyd: We're lucky that I noticed this before we did `0.16`, so there is no BC in the matter of `Style` class, so IMO we need to decide how to merge them (on what I'm open).

norberttech: ah wait, right, this is a new method that you just added 🤦‍♂️ so I would just move that StringStyles to `Flow\ETL\String\StringStyles` and make sure that the scalar function can use it and deprecated that other one used currently in ScalarFunction

norberttech (approved):
Looks good!

Issue #1500, opened 2025-02-26
Excel Adapter


norberttech:
Create an Excel Adapter based on: https://github.com/openspout/openspout/tree/4.x

Adapter should let us: 

- [ ] write to excel files (ideally append to specific sheets so we can create multi-pages sheets)
- [x] read from excel files

PR #1643, opened 2025-05-10
[Excel] Add a new Excel Adapter with an extractor

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new Excel Adapter with simple extractor</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


norberttech:
Shit, maybe reach out next time before you work on something, I hate to say it but I already took one attempt to build an Excel adapter (still have the codebase locally) and there is a roadmap item for it: https://github.com/flow-php/flow/issues/1500

There is one massive problem with Excel, it does not support remote filesystems :| 

The problem comes not really from openspout library but from ZipArchive that is used there under the hood. 
PHP implementation of ZipArchive makes it impossible to extract from archive in chunks. You need to pass entire zip archive at once, what means we can't read/write through byte range. 

I did some initial research and it is possible to create an implementation of ZipArchive that supports reading/writing from/to remote locations and I believe that openspout authors would have nothing against introducing reading/writing from/to remote locations: 

https://github.com/openspout/openspout/issues/120#issuecomment-1358906875

norberttech:
You can check my progress here in case you would like to investigate this one more: https://github.com/flow-php/flow/commit/87ebf9123bfecd8871a57ba5c4545e2b2fc3cf22

stloyd:
Remote support filesystem is nice, but I don't have use cases for remote filesystem, but a few with normal files.

This solves a particular problem, similar to yours (while mine does full reading), and solves a particular issue: reading an Excel file, which is IMO a valid use case even without support for remote filesystem.

I'm open to closing this and doing it in my project internally, but I guess that reading Excel files is quite a common case, which would be nice to cover.

norberttech:
> reading an Excel file, which is IMO a valid use case even without support for remote filesystem.

Totally, I meant more that if you would reach out before you could simply take over what I wrote and adjust it, saving yourself some typing. 

Let me think about and look at your code later, I will came up with a proper definition of done for this PR and then you can decide if you are willing to modify your code or drop it and implement internally. 

To your point, Excel Adapter is super valuable even without remote filesystem support and I would love to bring it to Flow. 

Although we could potentially support remote reading/writing by either doing it all in memory or buffering somewhere on the disk first before moving to remote storage - but that's out of the scope for now and not super critical as most of the use cases of this adapter are going to be anyway operating on local files only.

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+    private bool $withHeader = true;
+
+    public function __construct(private readonly Path $path)
+    {
+        $this->resetLimit();
+
+        $this->reader = match ($this->path->extension()) {
+            'xlsx' => new XlsxReader(),
+            'ods' => new OdsReader(),
+            default => throw new InvalidArgumentException('Unsupported file extension: ' . ($this->path->extension() ?: 'n/a')),
```
norberttech: Sometimes files might not have extension but they are still valid excel files and that wont work for them, what I would do here is introduce enum `ExcelReader` with two options `XLSX` and `ODS` and I would allow to set through `withReader(ExcelReader $type) : self` 

Then I would make a property: 

```
private null|XlsxReader|OdsReader $reader;
```

and a private method `private function reader() : XlsxReader|OdsReader` that would either: 

- create the one that user passed
- detect it from path extension
- throw an error when user set nothing and there is no extension

norberttech (changes requested):
As for definition of done for Extractors here is what I usually have in mind for new adapters: 

## Features
Below are the minimum features that should be supported by this extractor: 

- limit number of extracting rows
- partitioning 
- reading from xlsx/ods
- throw exception when path is not local fs 
- make it possible to read data only from specific sheet (excel files are in most cases comes with multiple sheets and the interesting one is usually not the first one) 
- support offset (reading from specific row), when withHeader is set to true, extractor should still read the first row but not consider is as a part of dataset. So when there are 10 rows and withHeader is true, we have only 9 rows. 

## Dependencies 

When adding new dependency make sure that it's added to the composer.json of the adapter and to the composer.json of the monorepo. Use monorepo builder tool (from tools) to validate it. 

As for the dependency version, try to support as wide range as possible. Try to not limit it jus to the latest version as many projects might not be able to upgrade and we still want them to use Flow. 

Obviously there needs to be some balance, so if too many things changed between versions then drop the support for older ones. But if it's about few methods being renamed or moved to another NS - we cover that. If you wont be sure how to tackle this, just leave a comment with what would need to happen to support version X and I will let you know if it's worth it or not (most likely not). 

Good example is Doctrine Dbal adapter where version 3.6 and 4 are supported. But again, that depends on how strict the external library is with semantic versioning. 

## Tests 

Make sure that tests are executed at CI/CD: `.github/workflows/test-suite.yml` 

### Options and Configuration
Each extractor should have at least one test for each of his options, so here with that reader type you should put a test will read a fixture file as xlsx and ods formats. 
Those tests should not only check how many rows are extracted but also check the values. 

### Remote Filesystem
There should be a test for using remote filesystems (when extractor does not support remote FS it should simply test that proper exception is thrown) 

### Partitioning 
All extractors must be able to extract from partitioned datasets. 

The test should not only cover read that throws no exception but should actually confirm that the data read from source is exactly how we expect it. This means that you need to prepare an excel files with let say 10 rows (or less), partition them manually (just create folders for partitions and put those files) and then read them and compare the results. 

### Limit 
There need to be a test proving that limiting behavior is working as expected and that when we set limit to 5 on 10 rows dataset we only get 5 rows. 

### Benchmarks 

Like most of the other extractors we should put benchmarks to monitor potentnial performance degradations. 
For that we need a dataset with 10k rows, for excel you might be able to create one by converting `src/adapter/etl-adapter-csv/tests/Flow/ETL/Adapter/CSV/Tests/Benchmark/Fixtures/orders_flow.csv` into excel file. 

## CLI 

New extractors that operate on files should also be added to the CLI, there are few commands that are critical: 

- src/cli/src/Flow/CLI/Command/FileReadCommand.php
- src/cli/src/Flow/CLI/Command/FileSchemaCommand.php
- src/cli/src/Flow/CLI/Command/FileRowsCountCommand.php 
- src/cli/src/Flow/CLI/Command/FileConvertCommand.php
- src/cli/src/Flow/CLI/Command/FileAnalyzeCommand.php

In this case you will need to also add `trait ExcelOptions` that would define options to pass into those commands. 

## Codebase 

Try to use dsl functions across the codebase, especially for types, rows, row, schema, extractors, loaders, dataframe, transformations, scalar functions. 

That applies also to tests (actually mostly to tests) 

## Documentation 

As you pointed out many times, docs and examples sucks right now (I'm perfectly aware and constantly trying to find time and motivation to improve it). 

Each new extractor should come with: 

### Examples 

You should add an example under `Data reading` topic. 

When extractor provides options they should be also documented in example description. (put a description.md file in the example folder) 

On top of that it would be good to put some explanation in the dsl function docblock as those functions are all part of the documentation later. 

Since this extractor can't read from remote filesystems it should also be mentioned in the example description and maybe also in dsl function docblocks. 

### Api References

Look at `phpdoc` folder and `build:docs:api` composer command. 

### Documentation 

If this is a new adapter, make sure it's available on the website: `web/landing/templates/documentation/navigation.html.twig` 

And that it comes with it's own readme file in `documentation/components/adapters` that is also listed in `documentation/installation.md`. 
In that readme file there should be a simple example about how to use that extractor with Flow. 


In order to generate commits stats new adapters/components needs to be also added to `manifest.json` as well as here: `.github/workflows/monorepo-split.yml`  

--- 

I believe that should cover most of it, I already created a repo for this adapter http://github.com/flow-php/etl-adapter-excel 

If I will think about anything else I will let you know as it's hard to write complete spec at one go but I really think that should be the most critical part. 

As for Loader it should also be added but this can happen later

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/DSL/functions.php
```diff
@@ -0,0 +1,27 @@
...
+
+#[DocumentationDSL(module: Module::EXCEL, type: DSLType::EXTRACTOR)]
+#[DocumentationExample(topic: 'data_reading', example: 'excel')]
+function from_excel(
+    string|Path $path,
+) : ExcelExtractor {
+    return new ExcelExtractor(\is_string($path) ? Path::realpath($path) : $path);
+}
+
+function valid_excel_sheet_name(string|ScalarFunction $sheet_name) : IsValidExcelSheetName
```
norberttech: is_valid_excel_sheet_name maybe?

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -0,0 +1,170 @@
...
+    private ?string $sheetName = null;
+
+    private bool $withHeader = true;
+
+    public function __construct(private readonly Path $path)
+    {
+        if (!$this->path->isLocal()) {
+            // We can't use resources (returned by \fopen) since they are not supported by the OpenSpout library.
+            // They are not supported because OpenSpout library uses php built in ZipArchive library, which doesn't support resources, only local paths.
+            throw new InvalidArgumentException('Only local filesystem paths are supported by ExcelExtractor due to the limitation of underlying library.');
```
norberttech, line 40: 👍

src/adapter/etl-adapter-excel/tests/Flow/ETL/Adapter/Excel/Tests/Unit/ExcelExtractorTest.php
```diff
@@ -0,0 +1,21 @@
...
+final class ExcelExtractorTest extends FlowTestCase
+{
+    public function test_invalid_sheet_name() : void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Sheet name must be a valid Excel sheet name');
+
+        from_excel(__DIR__ . '/../Fixtures/unknown')
+            ->withOnlySheet('This is veeeeeery long excel sheet name, longer than 32 characters');
+    }
```
norberttech, line 21: I guess there could be also a test to validate the exception thrown when path is not local

src/cli/src/Flow/CLI/Command/Traits/CSVOptions.php
```diff
@@ -29,4 +18,15 @@ private function addCSVInputOptions(Command $command) : void
             ->addOption('input-csv-enclosure', null, InputOption::VALUE_REQUIRED, 'CSV enclosure character')
             ->addOption('input-csv-escape', null, InputOption::VALUE_REQUIRED, 'CSV escape character');
     }
+
+    private function addCSVOutputOptions(Command $command) : void
```
norberttech, line 22: 👌

norberttech (commented):
I just wen through this PR from my phone, will do my best to review it later from a laptop but its a bit busy weekend so no promises.

In general it looks good, maybe I missed it but I haven't seen any tests for CLI commands that would cover excel input? (review from phone is not the most convenient so I might just miss them).

documentation/components/adapters/excel.md
```diff
@@ -0,0 +1,64 @@
...
+    ->fetch();
+```
+
+```php
+<?php
+
+$rows = data_frame()
+    ->read(
+        from_excel('path/to/your/excel.xlsx')
+            ->withOnlySheet('Sheet name')
```
norberttech: can we actually rename this to `->withSheetName()`? ->withOnlySheet looks a bit weird

norberttech (approved):

norberttech:
I think the code looks great, but there are two small things still missing: 

- test suite workflow needs to be modified to execute excel test suite
- monorepo-split needs to modified to split excel into https://github.com/flow-php/etl-adapter-excel

PR #1645, opened 2025-05-12
1628 make flow types fully compatible with static analysis tools

### LINKED ISSUE #1628: Make Flow Types fully compatible with Static Analysis Tools
Flow comes with it's own internal type system: 

- `type_float()` 
- `type_string()`
- etc 

The challange is that due to nullability being defined on a type level tools like phpstan can't properly narrow the type in methods like isValid(mixed $value) : bool 
like it does for tools like webmozart/assert for example. 

The goal is to also introduce `Type::assert(mixed $value) : mixed` so when we use it like this: 

```php
$floatValue = type_float()->assert($value); 
// $floatValue type is float, not float|null
```

the $floatValue is recognized as a proper `float` 

Additionally we also want to get the same behavior for 

```php
$floatValue = caster()->to(type_float()->value($value);
// $floatValue type is float, not float|null
```

norberttech:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>UnionType to Types</li>
    <li>OptionalType to Types</li>
    <li>nullable property to Schema\Definition</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Moved all Types related classes to flow-php/types component</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>nullable property from all types</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

Resolves: #1628


src/adapter/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/SchemaConverter.php
```diff
@@ -159,13 +159,13 @@ private function flowToColumn(string $name, Type $type, ?Schema\Metadata $metada
...
         $options = [
-            'notnull' => !$type->nullable(),
+            'notnull' => !$nullable,
         ];
 
         if ($type instanceof FloatType) {
             // with decimals precision and scale are confusing, in float precision is number of digits, not digits before/after decimal point
             // with decimals precision is total number of digits, and scale is number of digits after decimal point
-            $options['scale'] = $type->precision;
+            $options['scale'] = 6;
```
andrewprofile: I don't know the context but wasn't this change for testing purposes? If ok, ignore

src/cli/tests/Flow/CLI/Tests/Integration/FileSchemaCommandTest.php
```diff
@@ -90,57 +90,57 @@ public function test_run_schema_with_pretty_output() : void
...
         },
+        "nullable": false,
         "metadata": []
     },
     {
         "ref": "discount",
         "type": {
-            "type": "string",
-            "nullable": true
+            "type": "string"
```
andrewprofile, line 117: `string`? Should not be a `float`?

src/core/etl/src/Flow/ETL/PHP/Type/Logical/DateTimeType.php
```diff
@@ -4,102 +4,76 @@
...
 
-    public function nullable() : bool
-    {
-        return $this->nullable;
-    }
-
     public function toString() : string
     {
-        return ($this->nullable ? '?' : '') . 'datetime';
+        return 'datetime';
```
andrewprofile: I don't know the context where this was used but shouldn't the `type_optional` recognize it?

norberttech: Previously all types has `$nullable` property that why `date_type(true)` was returning `?datetime`. Now to achieve the same goal we need to do `type_optional(type_date())`

src/adapter/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/SchemaConverter.php
norberttech: Previously `type_float` had additional `precision` property that was used to round floats, which at the end of the day ended up with more complications so it was removed and rounding has become a user responsibility. 

I should remove it from here as well as scale can also be set by metadata 👍

src/cli/tests/Flow/CLI/Tests/Integration/FileSchemaCommandTest.php
norberttech: this particular test reads schema from a CSV file, without passing option `--auto-cast` to the command it will not try to guess the schema from the `shape of data` and when you read csv in php, everything is string: https://3v4l.org/7JSB5

PR #1659, opened 2025-05-19
Close Excel extractor when limit is reached

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Close Excel extractor when limit is reached</li>
    <li>Prevent memory leak in `ExcelExtractor` by not copying each row in loop</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -61,10 +61,12 @@ public function extract(FlowContext $context) : \Generator
 
         foreach ($context->streams()->list($this->path, $this->filter()) as $stream) {
             foreach ($this->extractRows($stream, $headers, $offset) as $row) {
-                $signal = yield array_to_rows($row, $context->entryFactory());
+                $signal = yield rows(array_to_row($row, $context->entryFactory(), $stream->path()->partitions()));
```
norberttech: it seems that there is a missing test here, no?

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -61,10 +61,12 @@ public function extract(FlowContext $context) : \Generator
...
                 $this->incrementReturnedRows();
 
                 if ($signal === Signal::STOP || $this->reachedLimit()) {
+                    $this->reader()->close();
```
norberttech: just out of curiosity, what happens when you don't close the reader?

norberttech: btw I just noticed that you also need to remove `Closure` from Extractor, Extractors don't need to implement Closures.

stloyd: It leaves the stream open to block access to the file, and if you do a few DF in a row, it may eat memory.

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
stloyd: I will try to generate a file that can address such an issue, it should be similar to the one used now when the last row has an empty cell.

norberttech: I was more refering to the fact that nothing is checking if partition is available at yielded Rows, which is pretty easy to reproduce in tests by testing reading from pratitioned dataset

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -162,6 +163,7 @@ private function extractRows(SourceStream $stream, array $headers, int $offset)
             foreach ($sheet->getRowIterator() as $rowIndex => $row) {
                 if (1 === $rowIndex && $this->withHeader) {
                     $headers = $this->createRowsFromCells($row);
+                    $previousRowDataCount = \count($headers);
```
stloyd: This will be reverted in favor of https://github.com/flow-php/flow/pull/1661

src/adapter/etl-adapter-excel/tests/Flow/ETL/Adapter/Excel/Tests/Integration/ExcelExtractorTest.php
```diff
@@ -175,4 +175,26 @@ public function test_extract_with_wrongly_selected_reader() : void
...
+                ['group' => 2, 'id' => 5, 'value' => 'e'],
+                ['group' => 2, 'id' => 6, 'value' => 'f'],
+                ['group' => 2, 'id' => 7, 'value' => 'g'],
+                ['group' => 2, 'id' => 8, 'value' => 'h'],
+            ],
+            df()
+                ->read(from_excel(__DIR__ . '/../Fixtures/partitioned/group=*/*.xlsx'))
+                ->withEntry('id', ref('id')->cast('int'))
+                ->sortBy(ref('id'))
+                ->fetch()
```
norberttech: you should firs just assign result of `->fetch()` to `$rows` variable and check if partition is available on it - that's what I meant in the comment about about missing test

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -171,6 +170,8 @@ private function extractRows(SourceStream $stream, array $headers, int $offset)
                 $rows[] = $rowData;
             }
 
+            $this->reader()->close();
+
             return $rows;
```
norberttech: I just noticed that, there is a memory issue here. 
Instead of of mergning each $row into `$rows[] = $rowData` you should `yield `$rowData` (ideally rename it to $row). 

Extractors by design needs to yield one row at time, this behavior can be adjusted by the users through `batchSize()` or `collect()`

stloyd: Good catch! Fixed in: [`57f557f` (#1659)](https://github.com/flow-php/flow/pull/1659/commits/57f557fcfa7000d56db4f7737c9e3fd428b82f62)

norberttech (approved):

PR #1661, opened 2025-05-19
Simplify Excel & GoogleSheet extractor cell expanding

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Simplify Excel & GoogleSheet extractor cell expanding</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>


src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
```diff
@@ -68,14 +68,9 @@ public function extract(FlowContext $context) : \Generator
...
-                    if ($headersCount > $rowDataCount) {
-                        \array_push(
-                            $rowData,
-                            ...\array_map(
-                                static fn (int $i) => null,
-                                \range(1, $headersCount - $rowDataCount)
-                            )
-                        );
+                    // Expand columns to the size of the previous row
+                    for ($i = $rowDataCount; $i < $headersCount; $i++) {
```
norberttech, line 72: what will happen in situation like this: 

- row 1 - 5 columns
- row 2 - 5 columns
- row 3 - 7 columns
- row 4 - 5 columns

the final will be like below? 

- row 1 - 5 columns
- row 2 - 5 columns
- row 3 - 7 columns
- row 4 - 7 columns

stloyd: I think right now: yes.

This will be similar case for Google (& Excel) sheets, when we can enforce drop of later extra cells or hardly fail.

norberttech: In that case I would skip that, users can simply use `df()->read()->select(...)` to make sure that only columns they need are present int the output and anything extra is removed, or they can use schema infer where those additional columns are going to be nullable and then they can pass inferred schema to extractor

norberttech (approved):

PR #1676, opened 2025-05-22
feat: Add UTF-8 BOM detection to CSV extractor

mkubasz:
- Add withBOMDetection() method to enable/disable BOM detection (enabled by default)
- Automatically skip UTF-8 BOM (0xEF, 0xBB, 0xBF) when reading CSV files
- Works with both local and remote files without modifying source
- Add integration tests for BOM handling with and without headers

<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

### DIFF
--- a/src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
+++ b/src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
@@ -32,6 +32,8 @@ final class CSVExtractor implements Extractor, FileExtractor, LimitableExtractor
 
     private bool $withHeader = true;
 
+    private bool $bomDetection = true;
+
     public function __construct(private readonly Path $path)
     {
         $this->resetLimit();
@@ -50,8 +52,17 @@ final class CSVExtractor implements Extractor, FileExtractor, LimitableExtractor
 
             $headers = [];
             $headersCount = 0;
+            
+            // Detect BOM before reading lines
+            $bomOffset = $this->detectBOM($stream);
+            $isFirstLine = true;
 
             foreach ($stream->readLines(length: $this->charactersReadInLine) as $csvLine) {
+                // Skip BOM on first line if present
+                if ($isFirstLine && $bomOffset > 0) {
+                    $csvLine = \substr($csvLine, $bomOffset);
+                }
+                $isFirstLine = false;
                 /** @var non-empty-list<null|string> $rowData */
                 $rowData = \str_getcsv($csvLine, $separator, $enclosure, $escape);
                 $rowDataCount = \count($rowData);
@@ -169,6 +180,13 @@ final class CSVExtractor implements Extractor, FileExtractor, LimitableExtractor
         return $this;
     }
 
+    public function withBOMDetection(bool $enabled = true) : self
+    {
+        $this->bomDetection = $enabled;
+
+        return $this;
+    }
+
     private function mapHeaders(array $headers) : array
     {
         $headers = \array_map(fn (string $header) : string => \trim($header), $headers);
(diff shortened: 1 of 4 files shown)


norberttech:
Nice, this one was much closer but still not close enough ;)

src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
```diff
@@ -50,8 +52,17 @@ public function extract(FlowContext $context) : \Generator
 
             $headers = [];
             $headersCount = 0;
+            
+            // Detect BOM before reading lines
+            $bomOffset = $this->detectBOM($stream);
+            $isFirstLine = true;
 
             foreach ($stream->readLines(length: $this->charactersReadInLine) as $csvLine) {
```
norberttech, line 60: For anyone reading this PR, the appropriate and much cleaner solution would be as follows: 

```php
foreach ($stream->readLines(length: $this->charactersReadInLine) as $line => $csvLine) {

    if ($line === 0 && $this->withBOMDetection && \str_starts_with($csvLine, "\xEF\xBB\xBF")) {
        $csvLine = \substr($csvLine, 3);
    }

```

Why is it a much better approach? Performance, performance, performance!

The solution proposed by AI introduces a new private function (which already looks like a code smell) that performs another stream read. So, we need to do the following:

1) Read 3 bytes from the stream to determine if we are reading with a BOM or not.
2) Regardless, read the exact same line using `$stream->readLines(length: $this->charactersReadInLine)`. 

For local filesystem files that might work, but... Those files might originate from a remote filesystem. This means that a CSV extractor for each file would need to make an additional HTTP request just to check if there's a BOM in the first line (which, anyway, will be read later). 

As always, the devil is in the details.

src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
```diff
@@ -184,4 +202,26 @@ private function mapHeaders(array $headers) : array
             \array_keys($headers)
         );
     }
+
+    /**
+     * Detects UTF-8 BOM at the beginning of the stream and returns the offset to skip it.
+     * 
+     * @return int The number of bytes to skip (3 if BOM is present, 0 otherwise)
+     */
+    private function detectBOM(SourceStream $stream) : int
```
norberttech, line 211: This method is just broken. It introduces unnecessary complexity. Besides reading the stream twice (which creates unnecessary HTTP requests for remote streams), it returns int. 
You might ask yourself why int? Well, it's an offset to skip when BOM is detected. 
But here is the twist: BOM can be only 3 bytes... So AI simply increased the complexity of our code.

Instead of returning the offset, that method (still unnecessary) should return bool to tell us if the given line starts from BOM or not.

src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
```diff
@@ -32,6 +32,8 @@ final class CSVExtractor implements Extractor, FileExtractor, LimitableExtractor
 
     private bool $withHeader = true;
 
+    private bool $bomDetection = true;
```
norberttech, line 35: It should be `$withBOMDetection` to keep the naming convention

src/adapter/etl-adapter-csv/tests/Flow/ETL/Adapter/CSV/Tests/Integration/CSVExtractorTest.php
```diff
@@ -99,6 +99,64 @@ public function test_extracting_csv_empty_headers() : void
         );
     }
 
+    public function test_extracting_csv_with_utf8_bom() : void
+    {
+        $extractor = from_csv(
+            __DIR__ . '/../Fixtures/csv_with_bom.csv'
+        );
+
```
norberttech, line 107: How can we even be sure that this fixture file starts from BOM? 

The more proper approach would be to put here an assertion that would read first 3 bytes from that fixture file and confirm that they are indeed BOM. 

(there is a test below that makes that assertion but... This creates a coupling between tests, through a fixture...)

PR #1679, opened 2025-05-24
Introduce `BatchingLoader` to improve `BatchSizeOptimization`

stloyd:
<h2>Change Log</h2>
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Introduce `BatchingLoader` to improve `BatchSizeOptimization`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    <li>Removed `BatchSizeOptimization::__construct($supportedLoaders)` parameter</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>
<hr/>

<h2>Description</h2>

With this change, we remove the requirement to update the `BatchSizeOptimization` class with new adapters, making DX easier. A developer needs to use the new interface in his/her loader.


norberttech:
So that's pretty much how I started building this feature. IT was based on a single interface but then I quickly realized this interface odes not provide any behavior (define any methods), it's just an empty shell which I don't like.

So I converted it to let the BatchOptimization holds the logic about which loaders should be optimized. That turned out to be a mistake, but I'm still very much against of adding empty interfaces. 

So here is the challange, we need to figure out a way to tell BatchOptmization that given instance of a Loader can be auto optimized (meaning that it can store data in batches and that this will improve general performance) but also what that batch size could be (right now it's hardcoded 1k). 

Maybe we could extend your proposition by creating a `BatchingLoader` interface that could live in `Loaders` namespace with `BatchingLoader::defaultBatchSize() : int` ?

src/core/etl/src/Flow/ETL/Loader/BatchingLoader.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Loader;
+
+interface BatchingLoader
+{
+    /**
+     * @return int<1, max>
+     */
```
norberttech, line 13: Lets describe that method purpose, something that will say the purpose of this interface is to set default batch size and that it's only applied when batch size is not explicitly set through `DataFrame::batchSize()`

src/core/etl/src/Flow/ETL/Pipeline/Optimizer/BatchSizeOptimization.php
```diff
@@ -27,26 +27,11 @@ final class BatchSizeOptimization implements Optimization
...
-        'Flow\ETL\Adapter\Doctrine\DbalLoader',
-        'Flow\ETL\Adapter\Elasticsearch\ElasticsearchPHP\ElasticsearchLoader',
-        'Flow\ETL\Adapter\Meilisearch\MeilisearchPHP\MeilisearchLoader',
-    ];
-
     /**
      * @param int<1, max> $batchSize
      */
-    public function __construct(private readonly int $batchSize = 1000, ?array $supportedLoaders = null)
+    public function __construct(private readonly int $batchSize = 1000)
```
norberttech: $batchSize is redundant, should come from the BatchingLoader method

stloyd: Yes, and no, we have a loader, but we have `Transformer` being allowed, so should we also have `BatchTransformer`, which is the same as the loader?

src/core/etl/src/Flow/ETL/Loader/BatchingLoader.php
stloyd: Added, have a look.

src/core/etl/src/Flow/ETL/Pipeline/Optimizer/BatchSizeOptimization.php
stloyd: Added new transformer as well.

norberttech: whats the use case for optimizing batch size for transformers? I'm not sure if I fully understand which transformer would benefit from transforming in batch?

stloyd: I don't have any, just passed element can be a transformer or loader, if that doesn't make sense I can remove that and enforce a check for loader only.

norberttech: yeah, I dont see how transformers could benefit from it, at least not now.

src/adapter/etl-adapter-meilisearch/src/Flow/ETL/Adapter/Meilisearch/MeilisearchPHP/MeilisearchLoader.php
```diff
@@ -4,32 +4,37 @@
...
+     * @param array{url: string, apiKey: string, batchSize?: int<1, max>, httpClient: ?ClientInterface} $config
      */
     public static function update(array $config, string $index) : self
     {
         return new self($config, $index);
     }
 
+    public function defaultBatchSize() : int
+    {
+        return $this->config['batchSize'] ?? 1000;
```
norberttech, line 35: Yeah I was worried this might happen. I'm honestly not sure if that's not worse than just keeping class names in the BatchOptmizaton :| 

So now, this interface creates an impression that batchSize should be controlled by the interface. 

So if you do: 

```php
df()
  ->read(...)
  ->batchSize(500)
  ->write(to_meilisearch()->withBatchSize(250))
  ->run()
```

It would be super confusing for someone who does not understand the purpose of batch size optimization.

Initially batch size was controlled by loaders, but it wasn't working well with DataFrame::batchSize(). This forced me to change the apporach and make all Loaders/Extractors to operate by default on a single row. 

For most loaders/extractors it doesn't matter but there are few that if the batch size is not set explicitly the performance will suffer.

For example DbalLoader - It's a massive difference between 1k inserts with 1 row and 1 insert with 1k rows. 
That's how I ended up with that BatchOptimization mechanism that was supposed to kick in only when user didn't set the batch size on the data frame.

norberttech:
After https://github.com/flow-php/flow/pull/1679/files#r2106259363 I'm honestly on the fence about this refactoring. On the one hand controlling which loaders should be auto optimized through an interface looks like a cleaner solution, but at the same time it creates new complexity. 
I'm afraid we need to came up with different solution but I'm afraid I don't have any better ideas for now, sorry :|

norberttech:
I spent some time thinking about it and here is what I came up with: https://github.com/flow-php/flow/issues/1688

It's not a final solution, more a proposal that should be discussed before anyone starts working on it. 

@stloyd your thoughts?

stloyd:
The description sounds reasonable, but I don't have time right now to give a proper review of the potential solution, as this one is a no-go, I will close it (it's still linked with the proposal from you so that it won't get lost).

PR #1685, opened 2025-05-27
feat: add bom removal in CSV extractor

### LINKED ISSUE #1669: CSV - BOM detection
As reported by Rajesh on X, BOM might be a bit problematic. Please read more about [BOM here](https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-with-bom). 

Image

BOM detection could happen here: https://github.com/flow-php/flow/blob/1.x/src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php#L45-L54

Ideally the reader should check for BOM and when it's present just skip and while reading the content. (It's important to remember that we should not modify the file as it might be remote readonly file) and instead we need to start reading from a different byte offset. 

As for the API I think this should be built in and automatically enabled, but there shuld be an option to disable BOM detecttion through: 

```
CSVExtractor::withBOMDetection(bool $enabled = true)
```

broncha:
Resolves: #1669

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add withBomRemoval() method to enable/disable BOM removal (enabled by default)</li>
  </ul> 
</div>


src/adapter/etl-adapter-csv/tests/Flow/ETL/Adapter/CSV/Tests/Integration/CSVExtractorTest.php
```diff
@@ -332,4 +332,124 @@ public function test_signal_stop() : void
         $generator->send(Signal::STOP);
         self::assertFalse($generator->valid());
     }
+
+    public function test_bom_removal_utf8(): void
+    {
+        $extractor = from_csv(
+            $path = Path::realpath(__DIR__ . '/../Fixtures/with_utf8_bom.csv'),
+        );
+
```
norberttech, line 469: could you maybe add here an assertion that will confirm that `__DIR__ . '/../Fixtures/with_utf8_bom.csv'` indeed starts with a correct BOM? I'm worried that at some point, someone might overwrite it even accidentally through the IDE and we will never notice that

norberttech:
I see that cs fixer is complaining, you can find explanation about how to prepare local development environment here: https://flow-php.com/documentation/contributing/environment/ (it also explains which commands to run locally before commiting)

norberttech (approved):
Thank you!!! 🙏

PR #1739, opened 2025-06-26
Improve performance of comparison transformations

norberttech:
Resolves: #xxx

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>performance of comparison transformations</li>
    <li>performance of calculating rows/nulls in parquet row groups</li>
    <li>make json and string types comparable</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>updated nix packages version in nix shell</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Function/Equals.php
+++ b/src/core/etl/src/Flow/ETL/Function/Equals.php
@@ -18,9 +18,11 @@ final class Equals extends ScalarFunctionChain
     public function eval(Row $row) : bool
     {
         $left = (new Parameter($this->left))->eval($row);
+        $leftType = (new Parameter($this->left))->asType($row);
         $right = (new Parameter($this->right))->eval($row);
+        $rightType = (new Parameter($this->right))->asType($row);
 
-        (new ValueComparator())->assertComparable($left, $right, '==');
+        (new ValueComparator())->assertComparableTypes($leftType, $rightType, '==');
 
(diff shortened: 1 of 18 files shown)


src/core/etl/src/Flow/ETL/Function/Equals.php
```diff
@@ -18,9 +18,11 @@ public function __construct(
     public function eval(Row $row) : bool
     {
         $left = (new Parameter($this->left))->eval($row);
+        $leftType = (new Parameter($this->left))->asType($row);
```
stloyd, line 21: If we get null as `$left` & `$right` should we still run version comparison?

norberttech: Not really, but it wont hurt, since the types comparator will get `type_null` 2x and will return that types are comparable. 
Adding extra condition just for nulles doesn't seem to bring any significant performance boost (was just testing this locally)

PR #1751, opened 2025-07-02
Skip double size counting on csv row normalizer

stloyd:
Reason: there is already a count of the size for the row at the beginning of the loop, so there is no point in recalculating it, while it cannot be modified in the middle of the loop.

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Skip double size counting on csv row normalizer</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
+++ b/src/adapter/etl-adapter-csv/src/Flow/ETL/Adapter/CSV/CSVExtractor.php
@@ -75,7 +75,7 @@ final class CSVExtractor implements Extractor, FileExtractor, LimitableExtractor
                     $headersCount = $rowDataCount;
                 }
 
-                $rowData = $rowNormalizer->normalize($rowData, $headersCount);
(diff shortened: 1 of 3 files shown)


norberttech:
According to our [Contributing Guidelines](https://flow-php.com/documentation/contributing/)  all PR's (except bug fixes) must be preceded by a [Proposal](https://github.com/flow-php/flow/issues/new?template=proposal.yml) so we can first review the idea before any code is written.

In this particular case I honestly don't see any significant performance improvements and it seems wrong to pass an array and array size next to it. So unless there is something I'm missing I would keep it as is. 
(that's why the contributing guidelines exists in the first place, so we can agree on what is needed or not before writing any code)

PR #1757, opened 2025-07-04
Bug parquet reader memory consumption

### LINKED ISSUE #1755: [Proposal]: Optimize Parquet Write Memory Consumption
### Describe the Proposal

Currently ParquetWritter is quite heavy on memory. The reason for that is related to how the writer works. 

Right now whenever we are writing rows ParquetWriter internally initializes a RowGroup, then it keeps adding to this Row Group until the row group size is greater or equal configured size. 
When it reaches the expected size it flushes the Row Group (writes it down). 

### API Adjustments

To improve memory consumption we need to take a bit different approach. 
Instead of holding entire RowGroup (Row Group has exactly one Column Chunk for each column and multiple Pages per column chunk) at once in memory we should hold only one Page at time in memory. 
Whenever page size is reached it should be flushed. 

This means that instead of keeping rows in memory we should aim to keep in memory only one page at time, and instead of holding in memory raw values we can pretty much write them to the binary buffer (which will also be compressed when we flush page). 

Then we should replicate the same behavior on the ColumChunk Builder level. 
So whenever we flush the page it should be written to a column chunk binary buffer and compressed. This means that we keep in memory only compressed pages before we reach the total row group size and flush all column chunks to file closing row group. 

Just by moving to writing data right away to binary buffers instead of keeping raw rows in memory will: 

- simplify checking if the page / row group size limit is reached as it will be simple strlen on buffer (or we can even keep a record of written bytes) 
- reduce significantly the amount of data we store in memory 

Which might be a really good first step

### Are you intenting to also work on proposed change?

Yes

None

_No response_

norberttech:
Resolves: #1755

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>high memory consumption even when reading small chunks</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

This is how it started: https://blackfire.io/profiles/fe7a4799-56bc-4cab-9171-31c3579d045b/graph
This is where we are now: https://blackfire.io/profiles/74a0e0b5-8d0c-4de3-a8a8-eb280a27c261/graph

image


stloyd:
For those who want to see that nice graph: https://app.blackfire.io/profiles/compare/fe7a4799-56bc-4cab-9171-31c3579d045b...74a0e0b5-8d0c-4de3-a8a8-eb280a27c261/graph?settings=%5Bobject+Object%5D&selected=null&callname=main%28%29&constraintDoc=null

norberttech:
This is just a first step to proper parquet memory optimization, next step would be to optimize writer https://github.com/flow-php/flow/issues/1755 and bring back size based control over row groups / pages

src/lib/parquet/src/Flow/Parquet/ParquetFile/RowGroupBuilder/PagesBuilder.php
```diff
@@ -40,8 +50,17 @@ public function build(FlatColumn $column, FlatColumnValues $data, ColumnChunkSta
...
         }
 
-        foreach ($data->splitByRows($this->pageSizeCalculator->rowsPerPage($column, $statistics)) as $rowsChunk) {
-            $containers->add((new DataPageBuilder($this->compression, $this->options))->build($column, $rowsChunk));
+        $rowsPerPage = $this->options->getInt(Option::PAGE_MAXIMUM_ROWS_COUNT);
+
+        // For nested columns, avoid page splitting to prevent sibling synchronization issues
+        if (\str_contains($column->flatPath(), '.')) {
+            // Nested column: use single page to ensure siblings stay synchronized
+            $containers->add((new DataPageBuilder($this->compression, $this->options))->build($column, $data));
```
norberttech: This is not a solution! Because of this this PR can't be merged, more explanation below: 

**Bug: Page synchronization breaks for deeply nested structures**

When writing nested Parquet structures with multiple pages, sibling columns end up with different page counts, causing read failures with "Unexpected child flat values" exception.

**Root cause:** The `splitByRows()` method counts logical rows (repetitionLevel=0), but different column types have different numbers of physical values per logical row:
- Simple fields (e.g., `address.street`): 1 value per logical row
- List fields (e.g., `notes.list.element`): Variable values per logical row
- Nested lists (e.g., `items.list.element.sku`): Even more variable

This causes page breaks at different absolute positions for each column type, even when counting the same number of "logical rows". During reading, the MultipleIterator expects all sibling columns to have the same number of pages, but they don't.

**Current workaround:** Force all nested columns to use a single page, ensuring synchronization at the cost of losing pagination benefits for large datasets.

**Proper fix would require:** Synchronizing page boundaries across all sibling columns based on their parent's logical row boundaries, not each column's individual row count.

---

This pretty much means that we need to decide when to flush page on a higher level, just like described in: #1755

norberttech: Done! I managed to rewrite most of the Writer and optimize it to keep as little values in memory as possible.
I still need to cleanup the codebase, improve tests coverage but the hardest part is done, it will be released with 0.21.0 in next 2 weeks.

norberttech:
This is the latest benchmark: https://app.blackfire.io/profiles/compare/fe7a4799-56bc-4cab-9171-31c3579d045b...d4cb1ab4-920e-44ce-a093-7b04f7637b60/graph?settings=%5Bobject+Object%5D&selected=null&callname=main%28%29&constraintDoc=null

image

Issue #696, opened 2023-10-31
Parquet allow user to specify column encoding


norberttech:
Currently, PagesBuilder is deciding what type of encoding applies to the columns, and we only support two types: 

- RLE_DICTIONARY
- PLAIN 

We should allow users to overwrite this, but keep in mind that Booleans should not allow for RLE_DICTIONARY encoding. 

I think encoding could be added to NestedColumn/FlatColumn static constructors and used later in PagesBuilder that should fallback to default logic when it's not specified.

PR #1764, opened 2025-07-13
Parquet - support for Delta Binary Pcked encoding

### LINKED ISSUE #697: Parquet - Implement DELTA encoding
https://parquet.apache.org/docs/file-format/data-pages/encodings/#a-namedeltaencadelta-encoding-delta_binary_packed--5

Supported Types: INT32, INT64

This encoding is adapted from the Binary packing described in [“Decoding billions of integers per second through vectorization”](https://arxiv.org/pdf/1209.2137v5.pdf) by D. Lemire and L. Boytsov.

In delta encoding we make use of variable length integers for storing various numbers (not the deltas themselves). For unsigned values, we use ULEB128, which is the unsigned version of LEB128 (https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128). For signed values, we use zigzag encoding (https://developers.google.com/protocol-buffers/docs/encoding#signed-integers) to map negative values to positive ones and apply ULEB128 on the result.

Delta encoding consists of a header followed by blocks of delta encoded values binary packed. Each block is made of miniblocks, each of them binary packed with its own bit width.

The header is defined as follows:

<block size in values> <number of miniblocks in a block> <total value count> <first value>
the block size is a multiple of 128; it is stored as a ULEB128 int
the miniblock count per block is a divisor of the block size such that their quotient, the number of values in a miniblock, is a multiple of 32; it is stored as a ULEB128 int
the total value count is stored as a ULEB128 int
the first value is stored as a zigzag ULEB128 int
Each block contains

<min delta> <list of bitwidths of miniblocks> <miniblocks>
the min delta is a zigzag ULEB128 int (we compute a minimum as we need positive integers for bit packing)
the bitwidth of each block is stored as a byte
each miniblock is a list of bit packed ints according to the bit width stored at the beginning of the block
To encode a block, we will:

Compute the differences between consecutive elements. For the first element in the block, use the last element in the previous block or, in the case of the first block, use the first value of the whole sequence, stored in the header.

Compute the frame of reference (the minimum of the deltas in the block). Subtract this min delta from all deltas in the block. This guarantees that all values are non-negative.

Encode the frame of reference (min delta) as a zigzag ULEB128 int followed by the bit widths of the miniblocks and the delta values (minus the min delta) bit-packed per miniblock.

Having multiple blocks allows us to adapt to changes in the data by changing the frame of reference (the min delta) which can result in smaller values after the subtraction which, again, means we can store them with a lower bit width.

If there are not enough values to fill the last miniblock, we pad the miniblock so that its length is always the number of values in a full miniblock multiplied by the bit width. The values of the padding bits should be zero, but readers must accept paddings consisting of arbitrary bits as well.

If, in the last block, less than <number of miniblocks in a block> miniblocks are needed to store the values, the bytes storing the bit widths of the unneeded miniblocks are still present, their value should be zero, but readers must accept arbitrary values as well. There are no additional padding bytes for the miniblock bodies though, as if their bit widths were 0 (regardless of the actual byte values). The reader knows when to stop reading by keeping track of the number of values read.

Subtractions in steps 1) and 2) may incur signed arithmetic overflow, and so will the corresponding additions when decoding. Overflow should be allowed and handled as wrapping around in 2’s complement notation so that the original values are correctly restituted. This may require explicit care in some programming languages (for example by doing all arithmetic in the unsigned domain).

The following examples use 8 as the block size to keep the examples short, but in real cases it would be invalid.

Example 1
1, 2, 3, 4, 5

After step 1), we compute the deltas as:

1, 1, 1, 1

The minimum delta is 1 and after step 2, the relative deltas become:

0, 0, 0, 0

The final encoded data is:

header: 8 (block size), 1 (miniblock count), 5 (value count), 1 (first value)

block: 1 (minimum delta), 0 (bitwidth), (no data needed for bitwidth 0)

Example 2
7, 5, 3, 1, 2, 3, 4, 5, the deltas would be

-2, -2, -2, 1, 1, 1, 1

The minimum is -2, so the relative deltas are:

0, 0, 0, 3, 3, 3, 3

The encoded data is

header: 8 (block size), 1 (miniblock count), 8 (value count), 7 (first value)

block: -2 (minimum delta), 2 (bitwidth), 00000011111111b (0,0,0,3,3,3,3 packed on 2 bits)

Characteristics
This encoding is similar to the [RLE/bit-packing](https://parquet.apache.org/docs/file-format/data-pages/encodings/#RLE) encoding. However the [RLE/bit-packing](https://parquet.apache.org/docs/file-format/data-pages/encodings/#RLE) encoding is specifically used when the range of ints is small over the entire page, as is true of repetition and definition levels. It uses a single bit width for the whole page. The delta encoding algorithm described above stores a bit width per miniblock and is less sensitive to variations in the size of encoded integers. It is also somewhat doing RLE encoding as a block containing all the same values will be bit packed to a zero bit width thus being only a header.

norberttech:
Resolves: #697 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Support for Delta Binary Packed encoding in Parquet</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

Delta Binary Packed is applicable to Int32 and Int64 (which means that it also applies to Timestamps and other logical types stored as int32/64). 

The goal is to significantly reduce output file size through storing only delta between integers in a column rather than integers themself. 

This is super useful when we are dealing with incremental datasets, like for example orders.


src/lib/parquet/src/Flow/Parquet/Writer/ColumnChunkBuilders.php
```diff
@@ -103,6 +104,10 @@ public function uncompressedSize() : int
 
     private static function createFlatColumnBuilder(FlatColumn $column, Options $options, Compressions $compressions) : ColumnChunkBuilder
     {
+        if ($column->type() === PhysicalType::INT32 || $column->type() === PhysicalType::INT64) {
```
norberttech: since Delta Binary Packed encoding was added in Parquet V2, we should use it only when V2 is set (or when it's requested by the user, but that requires #696 )

PR #1769, opened 2025-07-14
Allow to set custom compression algorithm for columns

### LINKED ISSUE #1766: [Proposal]: Allow to chose a specific compression for parquet columns
### Describe the Proposal

Right now we set the compression on a file level (which should stay as is) but on top of that we should let users to set different compressions on specific columns that would overwrite the global one. 

### API Adjustments

The idea is identical to changes from #1765 

### Are you intenting to also work on proposed change?

Yes

None

_No response_

norberttech:
Resolves: #1766

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Allow to set custom compression algorithm for columns</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/lib/parquet/src/Flow/Parquet/Options.php
```diff
@@ -45,37 +44,40 @@ public static function default() : self
     }
 
     /**
-     * @return null|array<mixed>|bool|ColumnsEncodings|float|int
+     * @return null|array<mixed>|bool|float|int
      */
-    public function get(Option $option) : bool|int|float|array|ColumnsEncodings|null
+    public function get(Option $option) : bool|int|float|array|null
```
norberttech, line 49: I'm reverting that change in favor of more generic approach

PR #1845, opened 2025-09-15
RetryLoader

### LINKED ISSUE #1842: [Proposal]: Retry Loader
### Describe the Proposal

Extractor/Loaders, especially those that relay on external services might crash from time to time, especially when we are dealing with networking for reasons unrelated to data. 

In those cases we would like to retry extracting/loading, with some small delay. 
However we should be able to have a flexible logic deciding if retry is even necessary. 

For  that we should create an `RetryStrategy` interface with few simple implementations: 

- OnException (allow to pass one or more exceptions classes, when non passed any Throwable triggers retry) 

Additionally we should be able to control the delay through `DelayFactory` interface with following implementations: 

- Fixed(Duration $delay) 
- Exponential(Duration $initial, int $multiplier) 
- Linear(Duration $delay, int $increment)
- Jitter(Duration $initial, float $randomness) 

**Duration** - this is a simple value object that have multiple static factories for fromMilliseconds / fromSeconds / fromMinutes (with private constructor) and that under the hood keeps everything as milliseconds and that return those milliseconds through method milliseconds(): int. 

**Sleep** - in order to not wait in tests, we also need to use `Sleep` abstraction. Simple interface that will let user to decide for how long process should sleep with 2 implementations. 
Sleep interface should also use TimeUnit value object `Sleep::for(Duration $duration)` 

- FakeSleep - for tests
- SystemSleep - for actual implementation that uses usleep 

**Namespaces**
- Loader should be implemented in following namespace: src/core/etl/src/Flow/ETL/Loader
- Extractor should be implemented in following namespace: src/core/etl/src/Flow/ETL/Extractor
- Sleep / Duration should be implemented in following namespace: src/core/etl/src/Flow/ETL/Time
- RetryStrategy / DelayFactory should be implemented in src/core/etl/src/Flow/ETL/Retry

### API Adjustments

DSL Functions

```
write_with_retries(
   Loader $loader,
   RetryStrategy $retryStrategy = new OnException(),
   DelayFactory $delayFactory = Fixed(TimeUnit::fromSeconds(1)),
   int $maxRetries = 3,
)
```

```
read_with_retries(
   Extractor $loader,
   RetryStrategy $retryStrategy = new OnException(),
   DelayFactory $delayFactory = new Fixed(TimeUnit::fromSeconds(1)),
   int $maxRetries = 3,
)
```

Yes

None

_No response_

norberttech:
Resolves: #1842

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>RetryLoader</li>
    <li>Retry mechanism building blocks</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/core/etl/src/Flow/ETL/Loader/RetryLoader.php
```diff
@@ -0,0 +1,46 @@
...
+
+            try {
+                $this->loader->load($rows, $context);
+
+                return;
+            } catch (\Throwable $exception) {
+                $retriesRecord->add(FailedRetry::create($context->config->clock(), $exception, $attemptNumber));
+
+                if (!$this->retryStrategy->shouldRetry($exception, $attemptNumber)) {
+                    throw new FailedRetryException($retriesRecord);
```
stloyd, line 39: IMO it should get previous exception to not lose stacktrace

norberttech: that's exactly how it works, each exception is added to retries record and at the end they are passed to FailedRetryException unless you meant something different?

stloyd: Right, missed that there is hold reference to the exception in the passed record.

PR #1917, opened 2025-10-27
[HTML] Add a new `HTMLType`

### LINKED ISSUE #1913: [Task]: HTML Type
### What needs to be done?

In order to complete #1912 we first need to add `type_html` to `flow-php/types` library that would follow the convention of other types allowing for assertions, casting and validation. 

`type_html()->toString()` should simply return `html` 

All requirements about types are documented here: 

- https://flow-php.com/documentation/components/libs/types/
- https://flow-php.com/documentation/components/libs/types/architecture/

stloyd:
Resolves: #1913 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `HTMLType`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/lib/types/src/Flow/Types/DSL/functions.php
```diff
@@ -425,6 +427,15 @@ function type_literal(bool|float|int|string $value) : LiteralType
     return new LiteralType($value);
 }
 
+/**
+ * @return Type<HTMLDocument>
+ */
+#[DocumentationDSL(module: Module::TYPES, type: DSLType::TYPE)]
+function type_html() : Type
```
norberttech, line 443: I think there should be a function here checking if HTLMDocument is available if not, throw an exception unless there is an existing polyfill that can replace non existin HTMLDocument class

src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Flow\Types\Type\Logical;
+
+use const Dom\HTML_NO_DEFAULT_NS;
+use function Flow\Types\DSL\type_string;
+use Dom\HTMLDocument;
+use Flow\Types\Exception\{CastingException, InvalidTypeException};
+use Flow\Types\Type;
+
+/**
+ * @implements Type<HTMLDocument>
```
norberttech, line 12: Hmm I'm not really sure how to handle this one, the only thing that comes to my mind is creating such class here if it doesn't exists to at least satisfy PHPStan. 

Like literally here above that class add `if (!class_exists(...) {}` and inside define HTMLDocument class stub with all the methods/constants we are using in the codebase

norberttech: Alternatively PHPStan has https://phpstan.org/user-guide/stub-files which might be useful here

src/lib/types/src/Flow/Types/DSL/functions.php
stloyd: I have moved back to old `\DOMDocument` as supporting both will be real pain for no particular gain in matter of the flow usage.

src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
stloyd: As above, moved to good old `\DOMDocument` instead.

norberttech:
Don't wanna be a party pooper here but I'm afraid \DOMDocument is not going to work for HTML's. 
Few reasons but probably the most important one is that you can use \DOM\ParentNode functions on it :|

src/lib/types/src/Flow/Types/Value/HTMLDocument.php
```diff
@@ -0,0 +1,59 @@
...
+use Flow\Types\Exception\InvalidArgumentException;
+
+final class HTMLDocument implements \Stringable
+{
+    private string $value;
+
+    public function __construct(string|object $value)
+    {
+        if (\is_string($value)) {
+            if (\class_exists(\Dom\HTMLDocument::class, false)) {
```
norberttech: shouldn't this be `'\Dom\HTMLDocument'` ?

src/lib/types/src/Flow/Types/Value/HTMLDocument.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static function fromString(string $value) : self
+    {
+        return new self($value);
+    }
+
+    public function __toString() : string
+    {
+        return '';
```
norberttech: this should return `$this-toString()`

src/lib/types/src/Flow/Types/Value/HTMLDocument.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\Types\Value;
+
+use Flow\Types\Exception\InvalidArgumentException;
+
+final class HTMLDocument implements \Stringable
```
norberttech, line 9: That whole model should be covered with extensive unit tests that would cover both scenarios, with and without HTMLDocument (ideally different set of tests based on different versions of PHP)

src/lib/types/src/Flow/Types/Type/Native/String/StringTypeChecker.php
```diff
@@ -108,6 +108,35 @@ public function isFloat() : bool
...
+    {
+        if ($this->string === '') {
+            return false;
+        }
+
+        if ('<' !== $this->string[0]) {
+            return false;
+        }
+
+        if (\preg_match('/(<!doctype(.+?)>)?<html(.+?)>(.+?)<\/html>/im', $this->string) === 1) {
```
norberttech: I think I would first check if HTMLDocument is available and try to use it, and when it fails then fallback to DOMDocument with this Regexp.

PR #1920, opened 2025-10-28
Simplify `AutoCaster`

stloyd:
While working on #1917, I have noticed that `AutoCaster` always gets pre-converted data (array instead of string json, `\DOMDocument` instead of string XML), so validating if the given value is scalar with logic type inside is not needed. This also removes the check for the array key, as it was never used outside of the loop.

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Simplify `AutoCaster`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/lib/types/src/Flow/Types/Type/AutoCaster.php
```diff
@@ -79,14 +68,6 @@ private function castToString(string $value) : mixed
             return type_boolean()->cast($value);
         }
 
-        if ($typeChecker->isJson()) {
-            return type_json()->cast($value);
-        }
-
-        if ($typeChecker->isUuid()) {
-            return type_uuid()->cast($value);
-        }
```
norberttech, line 88: I don't think those two can be removed, it seems that we are missing tests for  them but by removing those lines following code will fail, breaking backward compatibility of caster: 

```php
<?php

declare(strict_types=1);

use Flow\Types\Type\AutoCaster;
use Symfony\Component\Uid\Uuid;

require_once __DIR__ . '/vendor/autoload.php';

$output = (new AutoCaster())->cast(Uuid::v7()->toString());

dd($output); // this should be Uuid instance, not string 
```

src/lib/types/src/Flow/Types/Type/AutoCaster.php
```diff
@@ -36,24 +34,15 @@ public function cast(mixed $value) : mixed
      */
     private function castArray(array $value) : array
```
norberttech, line 35: Uhm I think I messed up a bit this `castArray`. 

The only thing it does now is that it narrows array types. 

For example when we have array `[1, 2, 3.0]` in which we have int's and float, it will narrow the type to `float` as all `int` can be represented by `float` however I don't think this should be responsibility of a class called `AutoCaster` 😅
Probably I was trying to fix some bug and I did it here without properly thinking this through 🤦‍♂️

I think this is much bigger issue that will need to be solved in separated PR (probably will require some more explicit array types narrowing mechanism) but for now removing key types makes sense.

norberttech: https://github.com/flow-php/flow/issues/1921

PR #1940, opened 2025-10-29
Add a new `HTMLEntry`

### LINKED ISSUE #1914: [Task]: HTML Entry
### What needs to be done?

Besides HTMLEntry that you added to the proposal we also need a schema Definition with HTML type. Just like we have for [XML](https://github.com/flow-php/flow/blob/1.x/src/core/etl/src/Flow/ETL/Schema/Definition.php#L210)

stloyd:
Resolves: #1914 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `HTMLEntry`</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


norberttech (approved):

norberttech:
there is a bug in https://github.com/flow-php/flow/blob/1.x/phpunit.xml.dist#L75C1-L80C17 (those should have lib- prefix) which prevents types tests from being executed on the CI/CD 🤦‍♂️ I noticed because on my local machine I have overwritten phpunjt.xml.dist and noticed that html type related tests are failing

norberttech:
(it was automatically closed since #1944 mentioned this as something that should be resolved)

norberttech:
I checked out locally to that branch (to figure out why infection is complaining) and when I execute: 

```
 nix-shell --arg php-version 8.2 --run "composer test:lib:types"
```

I'm getting: 

```
1) Flow\Types\Tests\Unit\Value\HTMLDocumentTest::test_create_with_invalid_html_on_old
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<p>invalid</p>'
+''

/Users/norbert/Workspace/flow-php/flow/src/lib/types/tests/Flow/Types/Tests/Unit/Value/HTMLDocumentTest.php:48
```

Which makes sense, that test is expecting that DOMDocument is going to wrap value with `<p></p>` but you added `\LIBXML_HTML_NOIMPLIED` that should prevent that and I have no idea why it's passing on GitHub Actions :| 

Different version of libxml perhaps? 🤔

norberttech:
As for CodeCove / Infection - this PR just does not have enough code coverage (I thought it's something else related to tools config). 

https://app.codecov.io/gh/flow-php/flow/pull/1940?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=flow-php

It adds some lines of code, but around 27% of HTMLEntry is not covered by any test

norberttech:
> Different version of libxml perhaps? 🤔

Yes, here is the result of the investigation around this issue: 

```
Breaking change: libxml2 2.14.0 (March 27, 2025)
The LIBXML_HTML_NOIMPLIED behavior changed fundamentally in libxml2 2.14.0 when the HTML parser was rewritten to use an HTML5-compliant tokenizer.
Key changes:
Before 2.14.0:

HTML parser based on HTML4 spec (underspecified)
LIBXML_HTML_NOIMPLIED worked leniently with plain text and fragments
Plain text like 'body' was tolerated even without proper HTML tags

From 2.14.0 onwards:

HTML tokenizer fully conforms to HTML5 specification
Tree construction still uses custom algorithm (not fully HTML5)
Stricter handling of invalid markup
LIBXML_HTML_NOIMPLIED now requires actual HTML elements - plain text without tags results in empty output
```

norberttech (changes requested):
HTMLEntry needs better coverage

norberttech:
After taking another look at `\Flow\Types\Value\HTMLDocument` I think there is a bigger problem with the implementation. 

`\Flow\Types\Value\Uuid` - when you pass string that is not Uuid or uuid string it will simply throw an exception. 

```php
    public function test_construct_with_invalid_string_uuid_throws_exception() : void
    {
        $this->expectException(InvalidArgumentException::class);
        new Uuid('invalid-uuid-string');
    }
```

`\Flow\Types\Value\HTMLDocument` is making some weird flips

```php
    #[RequiresPhp('>= 8.4')]
    public function test_create_with_proper_html_on_newer() : void
    {
        $html = '<html><body><div><span>bar</span></div></body></html>';
        $document = new HTMLDocument($html);

        self::assertSame($html, (string) $document);
    }
```

So what needs to happen is following: 

`HTMLDocument` needs to validate the input and check if passed value is either: 

`\DOM\HTMLDocument` - and extract string from it
`\DOMDocument` - and extract string from it 
`string` - nothing to extract 

Once values are extracted it should validate if given string is a valid HTML5 by checking if: 

* starts with `<!DOCTYPE html ...>`
* root element is named `html` 
* root element has exactly one `body`
* root element has exactly one `head`

So now the test that is failing because of different versions of libxml should simply throw an exception.

src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return $this->type;
+    }
+
+    public function value() : ?HTMLDocument
+    {
+        return $this->value;
+    }
+
+    public function withValue(mixed $value) : self
```
norberttech, line 115: this method should be covered in unit tests

src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
```diff
@@ -0,0 +1,119 @@
...
+        $this->metadata = $metadata ?: Metadata::empty();
+        $this->type = type_html();
+    }
+
+    public function __toString() : string
+    {
+        return $this->toString();
+    }
+
+    public function definition() : Definition
```
norberttech, line 49: this method should be covered in unit tests

src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
stloyd: Just FYI, no other entry has a test for this 😛

PR #1943, opened 2025-10-30
#1926 - change bechawior for entry factory 

### LINKED ISSUE #1926: [Bug]: Inconsistent Behavior of EntryFactory with Uuid
### What happened?

DataFactory::create() method is covering something that it shouldn't. 

If we want to create an Entry from a passed value without passing Type EntryFactory will use TypeDetector to figure out what type it is. 

`$valueType = (new TypeDetector())->detectType($value);` 

Then when returned type is string, it will also use StringTypeChecker to narrow the type based on string shape, so for example `2025-01-01` will be detected as `type_date`. 
That type is later passed to EntryFactory::createAs() method which creates entry. 

But there is additional condition in create method. 

`if ($valueType instanceof InstanceOfType) {` 

from which all conditions except the last one are redundant (TypeDetector covers them anyway). 

The problem starts here: 

```php
foreach (['Ramsey\Uuid\UuidInterface', Uuid::class, 'Symfony\Component\Uid\Uuid'] as $uuidClass) {
    if (\is_a($valueType->class, $uuidClass, true)) {
        $valueType = type_uuid();

        break;
    }
}
```

So what it does is very similar behavior to StringTypeChecker (which btw should probably be called StringTypeNarrower). Meaning that when it gets TypeInstanceOf (object) it will try to narrow the type to one of `'Ramsey\Uuid\UuidInterface', Uuid::class, 'Symfony\Component\Uid\Uuid'` (where Uuid::class is redundant, type checker covers that) 

### How to reproduce?

Remove that whole `if ($valueType instanceof InstanceOfType) {` from EntryFactory and run this [test](https://github.com/flow-php/flow/blob/5c7b37c367caccc126f6c59d1598181cea441e6f/src/adapter/etl-adapter-parquet/tests/Flow/ETL/Adapter/Parquet/Tests/Integration/ParquetTest.php#L70-L71)

_No response_

### Version

0.10.0+

### Relevant error output

```shell
Copy and paste any relevant error output (no backticks needed).
```

MrHDOLEK:
Resolves: #1926 

  <h2>Change Log</h2>
  <hr />
  <div id="change-log">
    <h4>Added</h4>
    <ul id="added">
      <li>TypeNarrower interface for type narrowing pattern</li>
      <li>StringTypeNarrower class implementing TypeNarrower interface</li>
      <li>InstanceOfTypeNarrower class for handling external UUID libraries</li>
      <li>Support for detecting additional string types: boolean, integer, float, date, datetime, timezone, null</li>
    </ul>
    <h4>Fixed</h4>
    <ul id="fixed">
      <li>Redundant InstanceOfType handling in EntryFactory::create() method</li>
      <li>Missing NullType support in EntryFactory::createAs() method</li>
    </ul>
    <h4>Changed</h4>
    <ul id="changed">
      <li>StringTypeChecker renamed to StringTypeNarrower with new interface contract</li>
      <li>Refactored EntryFactory to use modular TypeNarrower pattern</li>
      <li>EntryFactory now uses InstanceOfTypeNarrower for UUID library detection</li>
      <li>JsonType.isValid() method now uses StringTypeNarrower for validation</li>
    </ul>
    <h4>Removed</h4>
    <ul id="removed">
      <li>Duplicate UUID library detection logic from EntryFactory</li>
    </ul>
    <h4>Deprecated</h4>
    <ul id="deprecated">
      
    </ul>
    <h4>Security</h4>
    <ul id="security">
      
    </ul>
  </div>


src/lib/types/src/Flow/Types/Type/Native/String/StringTypeNarrower.php
```diff
@@ -0,0 +1,40 @@
...
+
+final readonly class StringTypeNarrower implements TypeNarrower
+{
+    public function narrow(Type $type, mixed $value) : Type
+    {
+        if (!$type instanceof StringType || !\is_string($value)) {
+            return $type;
+        }
+
+        $checker = new StringTypeChecker($value);
```
norberttech: The idea was to rename StringTypeChecker into `StringTypeNarrower` and apply the TypeNarrower on it

src/core/etl/src/Flow/ETL/Row/EntryFactory.php
```diff
@@ -223,6 +182,14 @@ public function createAs(string $entryName, mixed $value, Definition|Type $defin
                 return string_entry($entryName, type_optional(type_string())->cast($value), $metadata);
             }
 
+            if ($type instanceof HTMLType) {
+                return string_entry($entryName, type_optional(type_string())->cast($value), $metadata);
+            }
+
+            if ($type instanceof NullType) {
```
norberttech, line 173: I believe this is already handled above so null type shouldn't reach here but if I'm wrong, please cover this with test

src/core/etl/src/Flow/ETL/Row/EntryFactory.php
```diff
@@ -223,6 +182,14 @@ public function createAs(string $entryName, mixed $value, Definition|Type $defin
                 return string_entry($entryName, type_optional(type_string())->cast($value), $metadata);
             }
 
+            if ($type instanceof HTMLType) {
+                return string_entry($entryName, type_optional(type_string())->cast($value), $metadata);
```
norberttech: Please remove it, @stloyd is going to cover HTMLEntry when he is done with his part

src/core/etl/src/Flow/ETL/Row/EntryFactory.php
MrHDOLEK: In the case of processing rows that are always null, you will only pass when you specify the type; this can only be detected then. I added a test for this case.

src/core/etl/src/Flow/ETL/Row/EntryFactory.php
MrHDOLEK: fixed

src/lib/types/src/Flow/Types/Type/Native/String/StringTypeNarrower.php
MrHDOLEK: fixed

src/lib/types/src/Flow/Types/Type/Logical/InstanceOfTypeNarrower.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Flow\Types\Type\Logical;
+
+use function Flow\Types\DSL\{get_type, type_uuid};
+use Flow\Types\Type;
+use Flow\Types\Type\TypeNarrower;
+
+final readonly class InstanceOfTypeNarrower implements TypeNarrower
```
norberttech: This one requires unit tests just like StringTypeNarrower

src/lib/types/src/Flow/Types/Type/TypeNarrower.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Flow\Types\Type;
+
+use Flow\Types\Type;
+
+interface TypeNarrower
+{
+    /**
+     * @return Type<mixed>
+     */
+    public static function narrow(mixed $value) : Type;
```
norberttech: this should not be a static method

src/lib/types/src/Flow/Types/Type/Logical/InstanceOfTypeNarrower.php
MrHDOLEK: done

norberttech:
that's a solid piece of work! 💪
Even though it introduces small BC Break 
I believe more predictable EntryFactory is worth it

norberttech (approved):

PR #1948, opened 2025-11-01
Make `HTMLDocument` reliable

stloyd:
Resolves: #xxx

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Make `HTMLDocument` reliable</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
+++ b/src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Flow\Types\Type\Logical;
 
-use Flow\Types\Exception\{CastingException, InvalidTypeException};
+use Flow\Types\Exception\{CastingException, InvalidArgumentException, InvalidTypeException};
 use Flow\Types\Type;
 use Flow\Types\Value\HTMLDocument;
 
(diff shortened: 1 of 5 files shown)


norberttech:
It looks much easier now! 

What about scenarios where there are some spaces between html elements? Like: 

```html
<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>
```

You are removing `\n` and `\t` but will this regexp work for the above code?

stloyd:
While detecting `type_html` has no problems. When it comes to `HTMLDocument`, it will work only with a string in tests, cause otherwise it will be dependent on `libxml` when tested with objects... and that thing goes mad 😂 (it sometimes removes spaces, sometimes just a few of them ;))

<details>

```
Expected :'<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>'
Actual   :'<!DOCTYPE html><html><head><title></title></head>    <body><p>invalid</p>    </body></html>'
```
</details>

But I added a test for that case with a string.

norberttech:
> While detecting `type_html` has no problems. When it comes to `HTMLDocument`, it will work only with a string in tests, cause otherwise it will be dependent on `libxml` when tested with objects... and that thing goes mad 😂 (it sometimes removes spaces, sometimes just a few of them ;))
> 
> ```
> Expected :'<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>'
> Actual   :'<!DOCTYPE html><html><head><title></title></head>    <body><p>invalid</p>    </body></html>'
> ```
> 
> But I added a test for that case with a string.

As long as those regexps will detect `<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>` as valid html, I would not even test the output to make tests less fragile. 

All I would test is that following string: `<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>` is not throwing InvalidArgumentException

stloyd:
There is a test:
```php
    public function test_create_with_string_with_multiple_spaces() : void
    {
        $html = '<!DOCTYPE html><html>   <head><title></title></head>    <body><p>invalid</p>  </body>  </html>';

        $document = new HTMLDocument($html);

        self::assertSame(
            $html,
            $document->toString(),
        );
    }
```

PR #1959, opened 2025-11-05
#1955 - unify json entry

### LINKED ISSUE #1955: [Task]: Unify JsonEntry
### What needs to be done?

JsonEntry is not really following our convention, and definition of it looks like this: 

```php
/**
 * @implements Entry<?array<mixed>>
 */
final class JsonEntry implements Entry
```

Which assumes that json is an `array` but it should be `string`. 

Another problem is related to validation, which happens directly in the constructor, which is redundant (logic duplication), validation should use `type_json()->validate()`

MrHDOLEK:
Resolves: #1955

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/core/etl/src/Flow/ETL/DSL/functions.php
+++ b/src/core/etl/src/Flow/ETL/DSL/functions.php
@@ -545,6 +545,10 @@ function float_entry(string $name, float|int|string|null $value, ?Metadata $meta
 #[DocumentationDSL(module: Module::CORE, type: DSLType::ENTRY)]
 function json_entry(string $name, array|string|null $data, ?Metadata $metadata = null) : Entry
 {
+    if (\is_array($data)) {
+        return JsonEntry::object($name, $data, $metadata);
+    }
+
     return new JsonEntry($name, $data, $metadata);
 }
 
--- a/src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
+++ b/src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
@@ -5,14 +5,13 @@ declare(strict_types=1);
 namespace Flow\ETL\Row\Entry;
 
 use function Flow\Types\DSL\{type_equals, type_json, type_optional};
-use Flow\ArrayComparison\ArrayComparison;
 use Flow\ETL\Exception\InvalidArgumentException;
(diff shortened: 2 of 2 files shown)


src/core/etl/src/Flow/ETL/DSL/functions.php
```diff
@@ -545,6 +545,10 @@ function float_entry(string $name, float|int|string|null $value, ?Metadata $meta
 #[DocumentationDSL(module: Module::CORE, type: DSLType::ENTRY)]
 function json_entry(string $name, array|string|null $data, ?Metadata $metadata = null) : Entry
 {
+    if (\is_array($data)) {
```
norberttech, line 548: this doesn't make much sense, array can still be a regular array but not json object. 

```
{
   "id" : 1,
   "name": "test"
}
```

In PHP is `['id' => 1, "name" => "test"]` 

But `[{"id": 1, "name": "test"}]` in php would be `[["id":1, "name": "test"]]` however both satisfies php `is_array`. 

If anything I would probably do a `json_encode` on value when `is_array`

src/core/etl/src/Flow/ETL/Row/Entry/JsonEntry.php
```diff
@@ -136,15 +132,15 @@ public function isEqual(Entry $entry) : bool
                 && type_equals($this->type, $entry->type);
         }
 
-        return $this->is($entry->name()) && $entry instanceof self && type_equals($this->type, $entry->type) && (new ArrayComparison())->equals($thisValue, \is_array($entryValue) ? $entryValue : null);
+        return $this->is($entry->name()) && $entry instanceof self && type_equals($this->type, $entry->type) && $thisValue === $entryValue;
```
norberttech, line 135: something tells me this will break a lot of tests 🫢

PR #1960, opened 2025-11-06
Add `HTMLQuerySelectorAll` & `HTMLQuerySelector` scalar functions

### LINKED ISSUE #1915: [Task]: HTML Scalar Functions
### What needs to be done?

Ideally we should cover all functions from [DOMParentNode](https://www.php.net/manual/en/class.dom-parentnode.php) 

But we can start from: 

- `ScalarFunctionChain::querySelectorAll()` - returns all
- `ScalarFunctionChain::querySelector()` - returns first 

I don't think we need a dedicated DSL functions as those are type of functions that you execute on `ref('html')`

Like all other scalar functions those two would require: 

- Unit tests covering edge cases
- Simple integration tests covering happy path usage on some example pipeline

stloyd:
Resolves: #1915

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add `HTMLQuerySelectorAll` & `HTMLQuerySelector` scalar functions</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>`HTMLType & `HTMLEntry` require now `\Dom\HTMLDocument`</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


stloyd:
Coverage is a lie, as it runs on 8.2, but only part of the test is run on 8.4.

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
```diff
@@ -0,0 +1,35 @@
...
+    public function eval(Row $row) : mixed
+    {
+        $value = (new Parameter($this->value))->as($row, type_html());
+        $path = (new Parameter($this->path))->asString($row);
+
+        if (null === $value || null === $path) {
+            return null;
+        }
+
+        if (!\class_exists('\Dom\HTMLDocument')) {
```
norberttech: I think it would be safer to simply throw an exception here since it's not even about the mistake and fault tolerance, without this class query selector can't be even used, which means this function works only on PHP 8.4+

norberttech:
Both scalar functions must come with integration tests (like all other scalar functions) that shows how this works with a full pipeline

src/core/etl/tests/Flow/ETL/Tests/Integration/Function/HTMLQuerySelectorAllTest.php
```diff
@@ -0,0 +1,64 @@
...
+        self::assertCount(0, $results);
+    }
+
+    #[RequiresPhp('>= 8.4')]
+    public function test_valid_query_all_on_html_document() : void
+    {
+        $rows = df()
+            ->read(from_rows(rows(row(html_entry('html_raw', $this->html)))))
+            ->withEntry('html', ref('html_raw')->htmlQuerySelectorAll('body div span'))
+            ->drop('html_raw')
```
norberttech, line 40: what's the type of entry `html` here?

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
```diff
@@ -0,0 +1,35 @@
...
+            throw new \RuntimeException('This function requires \Dom\HTMLDocument extension available in PHP 8.4+.');
+        }
+
+        return HTMLDocument::createFromString($value->toString())
+            ->querySelector($path);
```
norberttech: also now when I think about, what's the purpose of having HTMLEntry if we anyway work on strings? :| We could have literally just use `string_entry`

I think we might be better with adding querySelector and querySelectorAll functions to HTMLDocument and make him to keep HTML as `\DOM\HTMLDocument` for `php >= 8.4` and as string for those below 8.4. 

Shit it's getting a bit complicated at this point due to 8.3 :|

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function __construct(
+        private readonly mixed $value,
+        private readonly ScalarFunction|string $selector,
+    ) {
+    }
+
+    public function eval(Row $row) : ?Element
+    {
+        if (\PHP_VERSION_ID < 80400) {
```
norberttech: the reasony why we always check if specific function/class exists instead of using hardcoded version constraints, is because when you check class/function polyfills can enable this. When you hardcode that it requires 8.4, even with polyfill this won't work. 

Please move towards checking if specific class/function/method exists every where

src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
```diff
@@ -24,24 +34,38 @@ public function assert(mixed $value) : HTMLDocument
...
+
+        if ('' === $value) {
+            return false;
+        }
+
+        if ('<' !== $value[0]) {
+            return false;
+        }
+
+        return \preg_match(self::HTML_ALIKE_REGEX, $value) === 1;
```
norberttech: all of the above is incorrect, TypeHTML expects `@implements Type<HTMLDocument>` - this means that string, even looking like html is not a valid HTMLDocument

src/lib/types/src/Flow/Types/Type/Native/String/StringTypeNarrower.php
```diff
@@ -33,71 +31,42 @@ public function narrow(mixed $value) : Type
...
-        }
-
-        if ($this->isInteger()) {
-            return type_integer();
-        }
-
-        if ($this->isTimeZone()) {
-            return type_time_zone();
+        if ($value === '') {
+            return type_string();
```
norberttech, line 38: any reason for that?

phpstan.neon
```diff
@@ -60,7 +60,15 @@ parameters:
...
         - src/core/etl/src/Flow/ETL/Formatter/ASCII/ASCIITable.php
+        - src/core/etl/src/Flow/ETL/Function/HTMLQuerySelectorAll.php
+        - src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
+        - src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
         - src/core/etl/src/Flow/ETL/Sort/ExternalSort/RowsMinHeap.php
+        - src/core/etl/tests/Flow/ETL/Tests/Integration/Function/HTMLQuerySelectorAllTest.php
+        - src/core/etl/tests/Flow/ETL/Tests/Integration/Function/HTMLQuerySelectorTest.php
+        - src/core/etl/tests/Flow/ETL/Tests/Unit/Function/HTMLQuerySelectorAllTest.php
+        - src/core/etl/tests/Flow/ETL/Tests/Unit/Function/HTMLQuerySelectorTest.php
+        - src/core/etl/tests/Flow/ETL/Tests/Unit/Row/Entry/HTMLEntryTest.php
```
norberttech: that's not great, can't we suppress only specific errors? Those paths are also prevent static analysis from doing other checks on those classes. 

+ if you keep it this way and we drop support for 8.2 and 8.3 we wont get even notified that those "exclusions" are not longer needed. 

When you suppress only specific errors, when there is nothing to suppress phpstan will notify us that given suppression is no longer needed.

src/lib/types/src/Flow/Types/Type/Native/String/StringTypeNarrower.php
stloyd: Every check in this narrower has done this assert internally, and at the end it was returning `type_string()`, but if it's an empty string, we can do that check once at the top.

src/lib/types/src/Flow/Types/Type/Logical/HTMLType.php
stloyd: Changed, moved the check to narrower.

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
stloyd: Adjusted, got confused as there was a mention to enforce version check.

phpstan.neon
stloyd: Maybe better to generate a baseline instead of any exclusion? This way, if anything changes in baseline, we will get note about "no error to ignore" report, WDYT?

stloyd: Here is a dedicated commit doing that: [`0156e72` (#1960)](https://github.com/flow-php/flow/pull/1960/commits/0156e720a9c9cf6d65e50ae9fa34faeb4bcd5878)

norberttech: The problem with a baseline is that I can't review it. Inline suppressions are much easier for me to take a look and say if dev is just lazy or if it's a true problem like here.

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelectorAll.php
```diff
@@ -0,0 +1,52 @@
...
+    ) {
+    }
+
+    /**
+     * @return null|array<Element>
+     */
+    public function eval(Row $row) : ?array
+    {
+        if (!\class_exists('\Dom\HTMLDocument')) {
+            throw new \RuntimeException('This function requires \Dom\HTMLDocument extension available in PHP 8.4+.');
```
norberttech: Pretty much everywhere in Flow we are using dedicated expcetions from Flow namespace. I believe we even have an exception for missing dependency, in this case I would create PHPVersionException

phpstan.neon
stloyd: Removed tests from exclusion, and left only dedicated classes.

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelectorAll.php
stloyd: Done.

phpstan.neon
```diff
@@ -60,6 +60,9 @@ parameters:
     excludePaths:
         - src/cli/src/Flow/CLI/Command/*
         - src/core/etl/src/Flow/ETL/Formatter/ASCII/ASCIITable.php
+        - src/core/etl/src/Flow/ETL/Function/HTMLQuerySelectorAll.php
+        - src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
+        - src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
```
norberttech: how many suppression those would require in file?
(I'm not home today, so sorry if I'm asking redundant questions, not everything is fully visible from phone 😅)

stloyd: Few are some not really fixable by comment as they are generics related that are saying given class doesn't exists 🫣

norberttech: Hmm doesn't phpstan allow to suppress specific issues in config for specific file/regex?

stloyd: It's called baseline which you don't like 🤣

norberttech: > It's called baseline which you don't like 🤣

It's called "knowing the tools you are using", more details in [documentation](https://phpstan.org/user-guide/ignoring-errors#ignoring-in-configuration-file) 

Errors related to missing \DOM\HTMLDocument can be suppressed like in the example below. 

```
parameters:
	ignoreErrors:
		-
			messages:
				- '#Call to an undefined method [a-zA-Z0-9\\_]+::doFooFoo\(\)#'
				- '#Call to an undefined method [a-zA-Z0-9\\_]+::doFooBar\(\)#'
			path: other/dir/AnotherFile.php
```

stloyd: I know that tool and what baseline is (which is a list of such ignored errors, nothing more). Done.

phpstan.neon
```diff
@@ -85,6 +85,61 @@ parameters:
         -
             function: 'dj()'
 
+    ignoreErrors:
+        -
+            message: '#^Cloning object of an unknown class Dom\\HTMLDocument\.$#'
```
norberttech: is message mandatory? 

How I was thinking about it, is something like that: 

```
        -
            identifier: class.notFound
            path: src/core/etl/src/Flow/ETL/Row/Entry/HTMLEntry.php
```

Since this is the only issue we really want to suppress. 

Then you also do the same for related scalar functions and move those suppressions here so after migrating fully to 8.4 there will be just one place to remove them all

stloyd: I understood that you want to exclude as much verbose as possible, but if not, then I adjusted to use a generic exclusion on HTML-related classes

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function __construct(
+        private readonly mixed $value,
+        private readonly ScalarFunction|string $selector,
+    ) {
+    }
+
+    public function eval(Row $row) : ?Element
+    {
+        if (!\class_exists('\Dom\HTMLDocument')) {
```
norberttech, line 17: could you actually move that to constructor? This way the exception is going to be thrown when you just use that function, otherwise it will crash only when pipeline is evaluated and only if it goes into this code.

norberttech (approved):

src/core/etl/src/Flow/ETL/Function/HTMLQuerySelector.php
stloyd: Here into the construct or more at `ScalarFunctionChain::htmlQuerySelectorAll()`?

norberttech: I would do it here since ScalarFunctionChain will pretty much call `new HTMLQueryselector` which will trigger exception

PR #1981, opened 2025-11-09
Add HTML usage example

### LINKED ISSUE #1916: [Task]: Prepare HTML Entry Examples
### What needs to be done?

Last step to complete new HTMLEntry should be adding some usage example. 

You can either use http adapter to grab some real html or simply prepare it from static string (static string is more reliable as it wont change over time). 

Would be good to add those related scalar function/functions to that example.  

The only challenge here is that we currently run examples on lowest supported version of PHP (.github/workflows/job-examples.yml - look here) 

What I would do, is to add to [run.php](https://github.com/flow-php/flow/blob/1.x/examples/run.php#L73-L77) a mechanism that would parse composer.json, check what's the minimum required php version for that given example and skip it (with a message) when current php version doesn't meet that expectation.

stloyd:
Resolves: #1916 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add HTML usage example</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


norberttech (approved):

norberttech:
I think that would be all, correct? 😁

stloyd:
To start real life testing for sure 😅

norberttech:
> To start real life testing for sure 😅

It will get released tomorrow with 0.27.0 🚀

PR #1994, opened 2025-11-10
Add a new `DOMElementParent` function

### LINKED ISSUE #1993: [Proposal]: Add a new `DOMElementParent` function
### Describe the Proposal

Using CSS selectors on the DOM is harder than using XPath, as you don't have access to the parent of the found element. But DOM elements hold this information internally as object relations:

```php
$document = HTMLDocument::createFromString('<article><section><h1>User Name</h1></section><span>01</span></article>');
$document->querySelector('section h1')->parentElement; // <section> relation
$document->querySelector('span')->parentElement; // <article> relation
```

### API Adjustments

```php
final class DOMElementParent extends ScalarFunctionChain
{
    public function __construct(
        private readonly ScalarFunction|\DOMNode|HTMLElement $element,
    ) {
    }

    public function eval(Row $row, FlowContext $context) : \DOMNode|HTMLElement|null
    {
         // ...

        if ($node instanceof \DOMNode || $node instanceof HTMLElement) {
            return $node->parentElement;
        }

        return null;
    }
}
```

Yes

None

_No response_

stloyd:
Resolves: #1993 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `DOMElementParent` function</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/core/etl/src/Flow/ETL/Function/DOMElementParent.php
```diff
@@ -0,0 +1,57 @@
...
+
+        if ($node instanceof HTMLElement) {
+            return $node->parentElement;
+        }
+
+        if ($node instanceof \DOMNode) {
+            return $node->parentNode;
+        }
+
+        return null;
```
norberttech: this should prbably also be handled through invalidResult as at this point it means that we have invalid type

src/core/etl/src/Flow/ETL/Function/DOMElementParent.php
```diff
@@ -0,0 +1,57 @@
...
+        if ($node instanceof \DOMDocument) {
+            $node = $node->documentElement;
+        }
+
+        if (\is_array($node) && \count($node)) {
+            $node = \reset($node);
+        }
+
+        if ($node === null) {
+            return $context->functions()->invalidResult(new InvalidArgumentException('DOMElementParent requires non-null DOMNode'));
```
norberttech: this should be covered with unit test under strict mode

src/core/etl/src/Flow/ETL/Function/DOMElementParent.php
stloyd: Covered.

src/core/etl/src/Flow/ETL/Function/DOMElementParent.php
stloyd: Done.

PR #2015, opened 2025-11-16
Enforcing string types for argument names `call()` scalar method

### LINKED ISSUE #2014: [Bug]: `call()` can lead to PHP fatal error when arguments are mixed
### What happened?

Mixing types of parameters will lead to a fatal error.

We should either:
1. Enforce check if parameter has numeric keys and `$refAlias` is an integer,
2. Enforce parameters to have only string keys and `$refAlias` to be a string,

Option 1 still has a problem from the second point, where the order of parameters will not match the function definition, either leading to invalid results or to deprecation warnings.

### How to reproduce?

```php
call('implode', ['separator' => ',']) // fatal
call('implode', [',']) // possible deprecation (wrong parameters order)
```

_No response_

### Version

0.27

### Relevant error output

```shell
In CallUserFunc.php line 50:
                                                                        
  Cannot use positional argument after named argument during unpacking
```

stloyd:
Resolves: #2014

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Enforcing string types for argument names `call()` scalar method</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/core/etl/src/Flow/ETL/Function/ScalarFunctionChain.php
```diff
@@ -152,10 +152,10 @@ public function binaryLength() : BinaryLength
     }
 
     /**
-     * @param array<array-key, mixed> $arguments
-     * @param Type<mixed> $returnType
+     * @param array<string, mixed> $arguments
+     * @param null|Type<mixed> $returnType
      */
-    public function call(ScalarFunction|callable $callable, array $arguments = [], string|int $refAlias = 0, ?Type $returnType = null) : CallUserFunc
+    public function call(ScalarFunction|callable $callable, string $refAlias, array $arguments = [], ?Type $returnType = null) : CallUserFunc
```
norberttech: I'm not sure how much I like that change. 
When `refAlias` is required string you need to always provide argument name to `call` even if function has just one parameter. 

For example `ref('array')->call('count')` needs to now become `ref('array')->call('count', refAlias: 'value')`

I just checked and in one project I have around 150 calls like this that I will need to manually one by one adjust (and sine those are different functions I would need to learn their argument names as well).
I need a really strong argument too approve that. 

Instead I would rather prefer to check if refAlias was passed as string|integer, then check if how arguments are passed and throw an exception if one us using integer indexes and other is string, or other way around.

stloyd: After writing more tests, I had a similar feeling that enforcing will always be problematic, and I did the change on it, I think the same way as you pointed out in the comment ;)

stloyd: By the way, which calls do you have more of? So we can extend the tests to see what I missed?

norberttech: The original intention of allowing for `int` was to be able to pass arguments by name or position. 

So for example: 

- `['param_name_1' => 'param_value_1', 'param_name_2' => 'param_value_2']`
- `['param_value_1', 'param_value_2']`

And ideally I would like to keep it this way. 

However to be able to use it with positional parameter you need to do something lke this:

```
call(ref('column'), ['param_1, null, 'param_2'], refALias: 1]) 
```

So array merge would replace `[1 => null]` with `[1 => ref('column')->eval()]`

(this wasn't properly documented)

PR #2095, opened 2025-12-20
bug: postgresql - clients types converting

norberttech:
Resolves: #xxx

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>added BoolArrayConverter for PostgreSQL BOOL[] type</li>
    <li>IntArrayConverter for PostgreSQL INT2[], INT4[], INT8[] types</li>
    <li>FloatArrayConverter for PostgreSQL FLOAT4[], FLOAT8[] types</li>
    <li>TextArrayConverter for PostgreSQL TEXT[], VARCHAR[] types</li>
    <li>UuidArrayConverter for PostgreSQL UUID[] type</li>
    <li>JsonArrayConverter for PostgreSQL JSON[], JSONB[] types</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>monolithic ArrayConverter in favor of type-specific converters</li>
    <li>syntax highlighing on website</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>array converters to throw ValueConversionException for invalid element types instead of silently converting</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
     <li>symfony/polyfill-83 dependency</li>
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


norberttech:
So this PR is pretty much removing flow-php/types dependency from flow-php/postgresql. 
Initially I wrongly assumed that we can based the php/postgresql types conversion through flow-php/types, however flow types are source, not destination types. 
Thats why passing flow type to `typed(mixed $value, Type $type)` didn't work as expected. 
What we really needed is passing PostgreSqlType as second argument (for which I created a dedicated DSL functions). 

On top of that array handling was terrible... Now all types from PostgreSqlType enum should be handled properly. 

We will need to add the most popular extensions as well, like PostGis for example and their types but that can come later. 

Last but not least, the data we are fetching from database is no longer automatically converted to any of PHP types that are ambiguous. 

This pretty much removes all "magic", fallbacking to exception rather than passing empty string, nulls, or automagically casted values

PR #2103, opened 2025-12-25
Simplify a little bit the `ExcelLoader` code

stloyd:
Resolves: #xxx

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Fixing bug in `ExcelExtractor` skipping the first row when no headers are needed</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Simplify a little bit the `ExcelLoader` code</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
+++ b/src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
@@ -52,10 +52,6 @@ final class ExcelExtractor implements Extractor, FileExtractor, LimitableExtract
         // Offset must be a positive number
         $offset = $this->offset ?? 1;
 
-        if (!$this->withHeader) {
-            $offset++;
-        }
-
         foreach ($context->streams()->list($this->path, $this->filter()) as $stream) {
             foreach ($this->extractRows($stream, $headers, $offset) as $row) {
                 // Ensure $row is an array before passing to array_to_rows
--- a/src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
+++ b/src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Flow\ETL\Adapter\Excel\RowsNormalizer;
 
+use Flow\ETL\Exception\InvalidArgumentException;
 use Flow\ETL\Row;
 use Flow\ETL\Row\Entry;
 use Flow\ETL\Row\Entry\{BooleanEntry,
@@ -11,14 +12,18 @@ use Flow\ETL\Row\Entry\{BooleanEntry,
     DateTimeEntry,
     EnumEntry,
     FloatEntry,
+    HTMLElementEntry,
+    HTMLEntry,
     IntegerEntry,
     JsonEntry,
     ListEntry,
     MapEntry,
     StringEntry,
     StructureEntry,
-    TimeEntry
-};
+    TimeEntry,
+    UuidEntry,
+    XMLElementEntry,
+    XMLEntry};
 
 final readonly class ExcelRowsNormalizer
 {
@@ -62,7 +67,7 @@ final readonly class ExcelRowsNormalizer
      */
     private function normalizeEntry(Entry $entry) : bool|float|int|string|null
     {
-        $value = match ($entry::class) {
+        return match ($entry::class) {
             BooleanEntry::class,
             IntegerEntry::class,
             FloatEntry::class,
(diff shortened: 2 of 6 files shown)


src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
```diff
@@ -11,25 +11,21 @@
...
 
 final readonly class ExcelRowsNormalizer
 {
     public function __construct(
-        private string $dateFormat = 'Y-m-d',
-        private string $dateTimeFormat = 'Y-m-d H:i:s',
-        private string $timeFormat = 'H:i:s',
+        private string $dateFormat,
+        private string $dateTimeFormat,
+        private string $timeFormat,
```
norberttech: I would prefer to keep default values here

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
```diff
@@ -74,16 +70,11 @@ private function normalizeEntry(Entry $entry) : bool|float|int|string|null
...
-            UuidEntry::class => $entry->toString(),
             EnumEntry::class => $this->normalizeEnumEntry($entry),
             JsonEntry::class,
             ListEntry::class,
             MapEntry::class,
             StructureEntry::class => $this->normalizeToJson($entry->value()),
-            XMLEntry::class,
-            XMLElementEntry::class,
-            HTMLEntry::class,
-            HTMLElementEntry::class => $entry->toString(),
```
norberttech, line 86: I don't think there is anything wrong with explicit conversion, maybe even exception on default would be a good idea in case we introduce some new entry type

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
```diff
@@ -52,10 +52,6 @@ public function extract(FlowContext $context) : \Generator
         // Offset must be a positive number
         $offset = $this->offset ?? 1;
 
-        if (!$this->withHeader) {
-            $offset++;
-        }
```
norberttech, line 57: that was actually a bug 🤦‍♂️

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/WorkbookManager.php
```diff
@@ -97,20 +100,18 @@ public function writeHeader(string $sheetName, array $headers, ?Style $style = n
...
+        if ($sheetName !== $writer->getCurrentSheet()->getName()) {
+            $sheet = $this->getOrCreateSheet($sheetName);
+            $writer->setCurrentSheet($sheet);
         }
 
-        $writer->setCurrentSheet($sheet);
+        $row = OpenSpoutRow::fromValuesWithStyles(
+            $values,
+            $styles ? \array_filter($styles, static fn (?Style $style) : bool => $style !== null) : []
+        );
```
norberttech, line 115:
```php
$writer->addRow(
  OpenSpoutRow::fromValuesWithStyles(
    $values,
    $styles ? \array_filter($styles, static fn (?Style $style) : bool => $style !== null) : []
  )
);
```

src/adapter/etl-adapter-excel/tests/Flow/ETL/Adapter/Excel/Tests/Unit/RowsNormalizer/ExcelRowsNormalizerTest.php
```diff
@@ -29,11 +29,22 @@ enum UnitTestEnum
 
 final class ExcelRowsNormalizerTest extends FlowTestCase
 {
-    public function test_headers() : void
+    private ExcelRowsNormalizer $normalizer;
```
norberttech: I'm usually trying to avoid state in test cases as much as possible

norberttech (approved):
just few minor comments

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
stloyd: The problem with new types is that they will be skipped anyway till added explicitly in any normalizer. The only workaround I see is to make a generic normalizer, with some explicit ones if needed for particular adapters. Cause i.e. we have already 
- csv,
- json,
- Parquet
- and now Excel
- potentially we can also have Google Sheets one,

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/ExcelExtractor.php
stloyd: Yes, it was missed, and introduced with the above fallback to `1` as offset, cause underlying library has an iterator iterating not from `0` (as it's normal in PHP) but from `1`...

src/adapter/etl-adapter-excel/src/Flow/ETL/Adapter/Excel/RowsNormalizer/ExcelRowsNormalizer.php
norberttech: they won't if there will be an exception on default case, no?

stloyd: If we fail hardly then yes, but still I think we can consider extracting a generic one ;)

norberttech: I took already at least 3 attempts to solve this problem 😂 The main challenge is that different destinations supports different structures, for example CSV requires nested structures serialized, meanwhile json not, parquet requires them to be kept as array etc 

Don't get me wrong, it would be awesome to at least extract the repetitive parts, but I'm having hard time finding them and making the abstraction easy to use/maintain

PR #2106, opened 2025-12-26
Unify adapter entry normalizers

stloyd:
Resolves: #xxx

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Unify adapter entry normalizers</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Row/Entry/EnumEntry.php
+++ b/src/core/etl/src/Flow/ETL/Row/Entry/EnumEntry.php
@@ -95,6 +95,10 @@ final class EnumEntry implements Entry
             return '';
         }
 
+        if ($this->value instanceof \BackedEnum) {
+            return (string) $this->value->value;
+        }
+
         return $this->value->name;
     }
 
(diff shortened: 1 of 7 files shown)


src/core/etl/src/Flow/ETL/Row/Entry/EnumEntry.php
```diff
@@ -95,6 +95,10 @@ public function toString() : string
             return '';
         }
 
+        if ($this->value instanceof \BackedEnum) {
+            return (string) $this->value->value;
+        }
+
```
stloyd, line 101: The only thing that, in theory, may be considered a BC break, but in practice, IMO, it makes sense to return a value for backed ones, not a name (the adapter does that already most of the time...).

norberttech:
sorry I'm not sure if I understand the reason behind that to be honest, and why it impacts the EnumEntry? (which seems to be a standalone problem here) 

Please use proposal process (unless asked otherwise - like when working on bigger thing split into smaller pieces) so I can at least understand the intention before any code is written

PR #2202, opened 2026-02-07
Symofny framework telemetry bridge

norberttech:
<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Flow PHP - Symfony Telemetry Bundle</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>DSL documentation page</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,380 @@
...
+                                ->info('Schema URL for semantic conventions')
+                                ->defaultNull()
+                            ->end()
+                            ->arrayNode('attributes')
+                                ->info('Additional scope attributes')
+                                ->normalizeKeys(false)
+                                ->useAttributeAsKey('name')
+                                ->prototype('variable')->end()
+                            ->end()
+                        ->end()
```
stloyd, line 330: Do we need some validation over those? Like if schema is URL-like etc?

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,380 @@
...
+                ->end()
+                ->arrayNode('telemetry')
+                    ->info('Auto-telemetry configuration')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('http_kernel')
+                            ->info('HTTP kernel request tracing configuration')
+                            ->canBeEnabled()
+                            ->children()
+                                ->arrayNode('exclude_routes')
```
stloyd: With routes, it's tricky to cover them well cause you sometimes want to exclude, i.e., 401/403/404, but you still want to have 200 covered

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/FlowTelemetryExtension.php
```diff
@@ -0,0 +1,915 @@
...
+
+            if ($customServiceId === null) {
+                throw new RuntimeException('service_id is required when transport type is "service"');
+            }
+            $container->setAlias($transportServiceId, $customServiceId);
+
+            return $transportServiceId;
+        }
+
+        $endpoint = $config['endpoint'] ?? 'http://localhost:4318';
```
stloyd: I think you enforce values on the configuration level, no?

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/Instrumentation/Messenger/TracingMiddleware.php
```diff
@@ -0,0 +1,77 @@
...
+        } finally {
+            $tracer->complete($span);
+        }
+    }
+
+    private function getShortClassName(string $className) : string
+    {
+        $parts = \explode('\\', $className);
+
+        return \end($parts);
```
stloyd, line 115:
```suggestion
        return \substr(\strrchr($className, '\\'), 1);
```

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
norberttech: I don't think so, all of those are being unit tested

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
norberttech: hmm what would be the reason to exclude 404's?

stloyd: Not often, but I had cases where 404 was triggered by some bots scanning IPs and kinda doing DDoS due to the amount of data being logged in a short time frame, but the data logged there was useless, as it was completely random data.

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
stloyd: I get it's unit tested, which is great, but I speak about DX here :) So validation for the developer using that bundle and passing random crap 😂

norberttech: I'm not sure I get it, the whole point of those attributes is to pass random crap to APM :P

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
norberttech: yea I see how this might be tempting, but it should be filtered by: 

- otel collector 
- apm itself 

This way the whole logic and cost of filtering is moved away from a process that we are collecting telemetry from keeping the impact minimal. 

I might need to even remove that regexp

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/Configuration.php
stloyd: Comments on mobiles doesn't work best ;) it was more related to schema urls or versions not attributes only ;)

norberttech: Those are just strings pretty much and thats how they are validated. If someone decides to pass 💩 emoji as version, well he will have shitty aggregation on APM side 😂

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/DependencyInjection/FlowTelemetryExtension.php
```diff
@@ -0,0 +1,915 @@
...
+        if (($config['messenger'] ?? false) && \interface_exists(MiddlewareInterface::class)) {
+            $definition = new Definition(TracingMiddleware::class);
+            $definition->setArgument(0, new Reference(Telemetry::class));
+            $container->setDefinition('flow.telemetry.messenger.middleware', $definition);
+        }
+
+        $twigConfig = $config['twig'] ?? [];
+
+        if ($twigConfig['enabled'] ?? false) {
+            if (!\class_exists(AbstractExtension::class)) {
```
bendavies, line 952: symfony has `\Symfony\Component\DependencyInjection\ContainerBuilder::willBeAvailable(...)` btw

look how framwork bundle uses it

norberttech: I was even thinking about using composer runtime InstalledVersion but this seems to solve it out of the box, nice!

norberttech: meh, I'm afraid it wont work with the monorepo :( 

In the monorepo setup:                                                                                                                                                                                                                                                                              
 * symfony/messenger and twig/twig are in require-dev (dev dependencies)                                                                                                                                                                                                                             
 * flow-php/symfony-telemetry-bundle is listed in the replace section, not as a regular installed package                                                                                                                                                                                            
 * The willBeAvailable method checks if the parent package (flow-php/symfony-telemetry-bundle) is installed as non-dev, but in a monorepo with replace, Composer doesn't register it the same way

src/bridge/symfony/telemetry-bundle/src/Flow/Bridge/Symfony/TelemetryBundle/Instrumentation/HttpKernel/HttpKernelSpanSubscriber.php
```diff
@@ -0,0 +1,193 @@
...
+
+        $tracer->complete($span);
+
+        $request->attributes->remove(self::SPAN_ATTRIBUTE);
+        $request->attributes->remove(self::TRACER_ATTRIBUTE);
+    }
+
+    private function matchesPattern(string $route, string $pattern) : bool
+    {
+        if (\str_starts_with($pattern, '/') && \str_ends_with($pattern, '/')) {
```
bendavies: if `/` is the only delimiter supported somewhere then you might want to say that.
infact that's the least likely one i'd use here, as i'd then have to escape slashes in in the url.
id be more likely to use `#` as a delimiter

bendavies: you are also excluding excluding the use of any regex modifiers here

bendavies: what about something like 
```php
private function matchesPattern(string $route, string $pattern): bool
{
    $result = @\preg_match($pattern, $route);

    if ($result !== false) {
        return (bool) $result;
    }

    return $route === $pattern;
}
```

norberttech: good point!

PR #2224, opened 2026-02-22
Rework `GoogleSheetExtractor` to use batch get functionality

### LINKED ISSUE #2223: [Bug]: `GoogleSheetExtractor` adapter can quite easily exceed rate limits
### What happened?

Due to the way `GoogleSheetExtractor` is built, it can quite easily exceed the default rate-limits set over the Google Sheet API, leading to a fatal error in the middle of pipeline extraction. This limit can be even easier to reach if your batch is small and the number of rows is large.

### How to reproduce?

Try to extract a big spreadsheet with ~100k rows.

### Playground snippet

_No response_

n/a

### Version

0.30+

### Relevant error output

```json
{
  "error": {
    "code": 429,
    "message": "Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute per user' of service 'sheets.googleapis.com' for consumer 'project_number:1089364807498'.",
    "errors": [
      {
        "message": "Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute per user' of service 'sheets.googleapis.com' for consumer 'project_number:1089364807498'.",
        "domain": "global",
        "reason": "rateLimitExceeded"
      }
    ],
    "status": "RESOURCE_EXHAUSTED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "RATE_LIMIT_EXCEEDED",
        "domain": "googleapis.com",
        "metadata": {
          "quota_unit": "1/min/{project}/{user}",
          "quota_location": "global",
          "service": "sheets.googleapis.com",
          "quota_limit": "ReadRequestsPerMinutePerUser",
          "consumer": "projects/1089364807498",
          "quota_metric": "sheets.googleapis.com/read_requests",
          "quota_limit_value": "60"
        }
      },
      {
        "@type": "type.googleapis.com/google.rpc.Help",
        "links": [
          {
            "description": "Request a higher quota limit.",
            "url": "https://cloud.google.com/docs/quotas/help/request_increase"
          }
        ]
      }
    ]
  }
}
```

stloyd:
Resolves: #2223

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Fixed wrong handling of skipping the header row in `GoogleSheetExtractor`</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Rework `GoogleSheetExtractor` to use batch get functionality</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
```diff
@@ -37,48 +37,60 @@ public function __construct(
...
         $shouldPutInputIntoRows = $context->config->shouldPutInputIntoRows();
 
-        while ([] !== $values) {
-            foreach ($values as $rowData) {
+        $headers = [];
+        $headersCount = 0;
+
+        $response = $this->service->spreadsheets_values->batchGet($this->spreadsheetId, array_merge($this->options, ['ranges' => $ranges]));
+
+        foreach ($response->getValueRanges() as $valueRange) {
```
stloyd, line 72: FYI, it's sorted by the ranges passed as request data, so no need to validate or think about mixed order (e.g., header coming after values).

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
```diff
@@ -37,48 +37,60 @@ public function __construct(
...
+        foreach ($spreadsheet->getSheets() as $sheet) {
+            if ($sheet->getProperties()->title === $this->columnRange->sheetName) {
+                $maxRows = $sheet->getProperties()->getGridProperties()->getRowCount();
 
                 break;
             }
         }
 
-        $headersCount = \count($headers);
+        $cellsRange = new SheetRange($this->columnRange, 1, $this->rowsPerPage, $maxRows);
```
stloyd, line 55: The maximum value of rows is needed for batch get calls, as asking out of range fails with a validation error, while a "normal" get call doesn't do such validation.

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
```diff
@@ -37,48 +37,60 @@ public function __construct(
...
-            foreach ($values as $rowData) {
+        $headers = [];
+        $headersCount = 0;
+
+        $response = $this->service->spreadsheets_values->batchGet($this->spreadsheetId, array_merge($this->options, ['ranges' => $ranges]));
+
+        foreach ($response->getValueRanges() as $valueRange) {
+            foreach ($valueRange->getValues() as $rowData) {
+                // Skip empty rows
+                if ([] === $rowData) {
```
norberttech: Before we also skipped empty rows?

norberttech (approved):

src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
stloyd: Yes:
```php
            foreach ($values as $index => $row) {
                if ([] === $row) {
                    // Remove empty rows at the beginning of a sheet
                    unset($values[$index]);

                    continue;
                }

stloyd: Hmmm, but in fact, this changes behaviour, it will now remove all empty rows. I will restore the previous to look only when checking headers.

norberttech: so in general I think that extractor logic should be as raw as possible, meaning that empty rows should be filtered out on data frame by the user, not extractor as one might actually need to extract empty rows to count them

stloyd: Yes, now it's the same as it was; it will only skip the first empty rows, not any of the others. The problem with leading empty rows at the beginning is that it will not allow the code to read, i.e., headers, as the code tries to find values.

PR #2237, opened 2026-03-01
Simplify code by extending variadic usage

### LINKED ISSUE #2236: [Proposal]: Additional array transformations when using partritioning or type comparison
### Describe the Proposal

I have noticed that when using partitioning, sorting, or type comparison, there are additional calls to array merge, array unpack, etc. This is done cause some methods have both normal variables and variadic ones.

I.e.:
```php
public function partitionBy(string|Reference $entry, string|Reference ...$entries) : self
```

But later on, those two are merged into one, which is not really needed, as a variadic variable can handle that well already.

### API Adjustments

```php
public function partitionBy(string|Reference ...$entries) : self
```

Yes

None

_No response_

stloyd:
Resolves: #2236

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Simplify code by extending variadic usage</li>
    <li>Don't allow creating an empty list of references</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


norberttech:
closing this one as it was covered by the ADR in #2241

PR #2240, opened 2026-03-01
Don't remove already existing file when doing partitioning

### LINKED ISSUE #2239: [Bug]: Filesystem partition saving run with overwrite removes all files not just the related ones
### What happened?

Right now, when having two data frames working on the same partitioning, and using the ~~native~~ any filesystem on the run of the second data frame, it will remove files from the previous one.

### How to reproduce?

```php
df()
    ->extract(/* */)
    ->partitionBy('order-year', 'order-month')
    ->mode(overwrite())
    ->load(to_csv(__DIR__.'/output/orders.csv'))
    ->run()
;

df()
    ->extract(/* */)
    ->partitionBy('order-year', 'order-month')
    ->mode(overwrite())
    ->load(to_csv(__DIR__.'/output/sales.csv'))
    ->run()
;
```

### Playground snippet

_No response_

If the bug requires specific data, provide it here:
- PHP array
- CSV
- JSON
- Other format if bug is format-specific

### Version

0.30+

### Relevant error output

```shell
Copy and paste any relevant error output (no backticks needed).
```

stloyd:
Resolves: #2239 

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Don't remove already existing file when doing partitioning</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/core/etl/src/Flow/ETL/Filesystem/FilesystemStreams.php
+++ b/src/core/etl/src/Flow/ETL/Filesystem/FilesystemStreams.php
@@ -41,7 +41,7 @@ final class FilesystemStreams implements \Countable, \IteratorAggregate
 
                     if ($this->saveMode === SaveMode::Overwrite) {
                         if ($fileStream->path()->partitions()->count()) {
-                            $partitionFilesPatter = \Flow\Filesystem\DSL\path($fileStream->path()->parentDirectory()->uri() . '/*', $fileStream->path()->options());
+                            $partitionFilesPatter = Path::from($fileStream->path()->parentDirectory()->uri() . '/' . $fileStream->path()->basename(), $fileStream->path()->options());
 
                             foreach ($fs->list($partitionFilesPatter) as $partitionFile) {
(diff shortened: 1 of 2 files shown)


norberttech:
honestly I'm on a fence with this one, on the one hand, it seems to be mergable, but if we merge it, we are creating a rule that partitioning with overwrite mode should only remove the expected partition file - and I'm not sure how this will impact parallel processing. 

So if this is really problematic for your use case, maybe lets take a look together at your use case so I can get better understanding of what you are trying to achieve as there might be a different solution

PR #2242, opened 2026-03-03
Prevent fatal error when fetching Google sheet rows without values

### LINKED ISSUE #2243: [Bug]: Google sheet extractor fails when reading in batch empty rows
### What happened?

When reading a Google sheet with a lot of additional rows (by default sheet has 1k rows), saving it with a larger amount of data, like 1,5k and removing part of it, the spreadsheet metadata will hold the value as the maximum row count.

Docs say it should always be an array, but it can be `null`:
```
  /**
   * @return array[]
   */
  public function getValues()
```

### How to reproduce?

```php
  0 => Google\Service\Sheets\ValueRange {
    #internal_gapi_mappings: []
    #modelData: []
    #processed: []
    #collection_key: "values"
    +majorDimension: "ROWS"
    +range: "Sheet1!A1:Z1000"
    +values: array:898 []
},
1 => Google\Service\Sheets\ValueRange {
    #internal_gapi_mappings: []
    #modelData: []
    #processed: []
    #collection_key: "values"
    +majorDimension: "ROWS"
    +range: "Sheet1!A1001:Z1801"
    +values: null // here we have the reason of the fatal error
  }
```

### Playground snippet

_No response_

Google sheet

### Version

0.31+

### Relevant error output

```shell
In GoogleSheetExtractor.php line 73:
                                                                        
  [ErrorException]                                                      
  Warning: foreach() argument must be of type array|object, null given  
                                                                        

Exception trace:
  at /project/vendor/flow-php/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php:73
 Flow\ETL\Adapter\GoogleSheet\GoogleSheetExtractor->extract() at n/a:n/a
```

stloyd:
Resolves: #2243

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    <li>Prevent fatal error when fetching Google sheet rows without values</li>
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
+++ b/src/adapter/etl-adapter-google-sheet/src/Flow/ETL/Adapter/GoogleSheet/GoogleSheetExtractor.php
@@ -70,7 +70,7 @@ final class GoogleSheetExtractor implements Extractor, LimitableExtractor
         $response = $this->service->spreadsheets_values->batchGet($this->spreadsheetId, array_merge($this->options, ['ranges' => $ranges]));
 
         foreach ($response->getValueRanges() as $valueRange) {
-            foreach ($valueRange->getValues() as $rowData) {
+            foreach ($valueRange->getValues() ?: [] as $rowData) {
                 $rowDataCount = \count($rowData);
 
                 if ($this->withHeader) {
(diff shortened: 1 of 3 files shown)


norberttech:
It would be good to have a test for that

norberttech (approved):

PR #2268, opened 2026-03-24
feature: flow-php/arrow-ext

norberttech:
<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>arrow-ext and expose parquet reader/writer</li>
    <li>ParquetEngine abstraction into flow-php/parquet</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Let flow-php/parquet use engines based on arrow extension availability</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

**Heads Up** - I'm not rust developer and I don't have solid rust experience, building this extension was a learning process inspired by @ptondereau

Reviews, especially the rust part are welcome! 

To use extension in nix shell just use `nix-shell` which has argument `with-arrow-ext` set by default to `true`. 
To develop the extension (compile it and run tests) use `nix-shell --arg with-rust=true`


src/extension/arrow-ext/src/parquet/type_converter.rs
```diff
@@ -0,0 +1,1415 @@
...
+}
+
+macro_rules! build_int_array {
+    ($values:expr, $column_name:expr, $builder_type:ty, $cast_type:ty) => {{
+        let mut builder = <$builder_type>::with_capacity($values.len());
+        for zv in $values {
+            if zv.is_null() {
+                builder.append_null();
+            } else if let Some(v) = zv.long() {
+                builder.append_value(v as $cast_type);
```
ptondereau: this can lead to integer truncation here because ext-php-rs doesn't check internally to type.
You should use try_from like:

```rust
let converted = <$cast_type>::try_from(v).map_err(|_| {
                      format!(
                          "Column '{}': value {} out of range for {} ({}..={})",
                          $column_name,
                          v,
                          stringify!($cast_type),
                          <$cast_type>::MIN,
                          <$cast_type>::MAX,
                      )
                  })?;
                  builder.append_value(converted);
```

src/extension/arrow-ext/src/parquet/type_converter.rs
```diff
@@ -0,0 +1,1415 @@
...
+                } else {
+                    return Err(format!(
+                        "Column '{}': expected int or string, got {}",
+                        column_name,
+                        zval_type_name(zv)
+                    ));
+                }
+            }
+            Ok(Arc::new(builder.finish()) as ArrayRef)
+        }
```
ptondereau, line 1044: same here 
```suggestion
        DataType::UInt64 => {
	      let mut builder = UInt64Builder::with_capacity(values.len());
	      for zv in values {
	          if zv.is_null() {
	              builder.append_null();
	          } else if let Some(v) = zv.long() {
	              let converted = u64::try_from(v).map_err(|_| {
	                  format!(
	                      "Column '{}': value {} out of range for u64 (cannot be negative)",
	                      column_name, v,
	                  )
	              })?;
	              builder.append_value(converted);
	          } else if let Some(s) = zv.str() {
	              let parsed: u64 = s.parse().map_err(|e| {
	                  format!("Column '{}': failed to parse '{}' as u64: {}", column_name, s, e)
	              })?;
	              builder.append_value(parsed);
	          } else {
	              return Err(format!(
	                  "Column '{}': expected int or string, got {}",
	                  column_name,
	                  zval_type_name(zv)
	              ));
	          }
	      }
	      Ok(Arc::new(builder.finish()) as ArrayRef)
	  }
```

src/extension/arrow-ext/src/parquet/type_converter.rs
```diff
@@ -0,0 +1,1415 @@
...
+}
+
+macro_rules! build_float_array {
+    ($values:expr, $column_name:expr, $builder_type:ty, $cast_type:ty) => {{
+        let mut builder = <$builder_type>::with_capacity($values.len());
+        for zv in $values {
+            if zv.is_null() {
+                builder.append_null();
+            } else if let Some(v) = zv.double() {
+                builder.append_value(v as $cast_type);
```
ptondereau: same here:
```rust
let converted = v as $cast_type;
if converted.is_infinite() && !v.is_infinite() {
    return Err(format!(
        "Column '{}': value {} out of range for {}",
        $column_name, v, stringify!($cast_type),
    ));
}
builder.append_value(converted);
```

src/extension/arrow-ext/src/stream/php_source_stream.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+impl ChunkReader for PhpSourceStream {
+    type T = Cursor<Vec<u8>>;
+
+    fn get_read(&self, start: u64) -> parquet::errors::Result<Self::T> {
+        let length = self.inner.len.saturating_sub(start) as usize;
+        let bytes = self.get_bytes(start, length)?;
+        Ok(Cursor::new(bytes.to_vec()))
```
ptondereau: there is double copy by using .to_vec()

```suggestion
        Ok(Cursor::new(bytes.into()))
```

and if parquet crate support buffer in Read you can `bytes::Buf::reader()`

src/extension/arrow-ext/src/stream/php_destination_stream.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl Write for PhpDestinationStream {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let mut data = Zval::new();
+        data.set_binary(buf.to_vec());
+
+        let _result = self.obj
+            .try_call_method("append", vec![&data as &dyn IntoZvalDyn])
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("Failed to call append(): {:?}", e)))?;
+
```
ptondereau, line 40: If the PHP append() writes fewer bytes than expected (partial write, broken pipe), the parquet writer thinks everything succeeded. This can produce corrupted parquet files.

```rust
let result = self.obj.try_call_method("append", ...)?;
if let Some(written) = result.long() {
    Ok(written as usize)
} else {
    Ok(buf.len())  // fallback for void return
}
```

.github/workflows/job-arrow-extension.yml
```diff
@@ -0,0 +1,84 @@
...
+
+jobs:
+  build:
+    name: Build and Test Extension
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+        php: ['8.3', '8.4', '8.5']
```
ptondereau, line 17: worth testing ZTS?

src/extension/arrow-ext/composer.json
```diff
@@ -0,0 +1,34 @@
...
+    "extra": {
+        "branch-alias": {
+            "dev-1.x": "1.x-dev"
+        }
+    },
+    "php-ext": {
+        "extension-name": "arrow",
+        "build-path": "ext",
+        "configure-options": [],
+        "support-zts": false,
```
ptondereau: What is the rationale for not supporting ZTS?

src/extension/arrow-ext/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[global_allocator]
+static GLOBAL: System = System;
+
+#[php_module]
+pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
+    module
+        .class::<Reader>()
+        .class::<Writer>()
+}
```
ptondereau, line 30: A good practice is to show in the module table your extension, version etc..
```rust
pub extern "C" fn php_module_info(_module: *mut ModuleEntry) {
    info_table_start!();
    info_table_row!("flowphp/arrow", "enabled");
    info_table_row!("parquet version", env!("CARGO_PKG_VERSION"));
    info_table_end!();
}
```
and register it with `module.info_function(php_module_info)`

src/extension/arrow-ext/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+mod parquet;
```
ptondereau, line 1: if you want to support windows, you can add these features gate:

`#![cfg_attr(windows, feature(abi_vectorcall))]`

ptondereau (commented):
This is mostly good here! Great work!
I left some comments. Please don't hesitate to acknowledge

src/extension/arrow-ext/src/stream/php_destination_stream.rs
norberttech: good point but I think not on the rust level but php level, the php stream must guarantee that each append is either full or throw an exception, I'm going to fix it on the flow filesystem level

src/extension/arrow-ext/src/lib.rs
norberttech: nah, I gave up on Windows (partially), Flow does provide a basic Windows support (not entire test suite is ready for it) but that's all I'm willing to spend time on :P

src/extension/arrow-ext/src/stream/php_destination_stream.rs
norberttech: done, now those assertions (if all bytes were written) happens on the flow filesystem level 💪

src/extension/arrow-ext/src/lib.rs
norberttech: that actually points out a different problem - what's the rust way of making sure that Cargo.toml version actually composer from the git tag/commit? 

Like for now it's hardcoded 0.1.0 - but in reality the version of this library is coupled with the Flow releases, where every 2 weeks I bump the tag, so it's 0.33.0 now and thats what Cargo.toml should reflect but there is no way I will remember to manually change that version

src/extension/arrow-ext/composer.json
norberttech: I skipped ZST initially as I had issue with it, should work just fine now

src/extension/arrow-ext/src/lib.rs
ptondereau: You can actually begin to 0.33.0 or use a build.rs to compute at compile time the real version for another support

norberttech: yea it must be aligned with the monorepo, you have any examples of how others are doing the version setup during compilation time? Is it just a simple rust script that updates Cargo.toml?

ptondereau: there is a whole doc here https://doc.rust-lang.org/cargo/reference/build-scripts.html
If you need some deps just for this script, you can use https://doc.rust-lang.org/cargo/reference/build-scripts.html#build-dependencies

norberttech:
@ptondereau I believe it should be good now, thanks for your help! If you want to take another look, I'm probably going to aim to merge this one tomorrow

ptondereau:
LGTM!

PR #2313, opened 2026-04-18
Postgresql mapping context

norberttech:
<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Context to PostgreSql RowMapper</li>
    <li>StaticFactory row mapper implementation for PostgreSql Client</li>
    <li>Custom context to PostgreSql Client - supported by Symfony Bundle</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/lib/postgresql/tests/Flow/PostgreSql/Tests/Mother/MapperContextMother.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public static function with(
+        ?Query $query = null,
+        ?Client $client = null,
+        ?ClientContext $clientContext = null,
+    ) : Context {
+        return new Context(
+            $query ?? new Query('SELECT 1'),
+            $client ?? new StubClient(),
+            $clientContext ?? new ClientContext(),
```
stloyd, line 29:
```suggestion
        Query $query = new Query('SELECT 1'),
        Client $client = new StubClient(),
        ClientContext $clientContext = new ClientContext(),
    ) : Context {
        return new Context(
            $query,
            $client,
            $clientContext,
```

src/lib/postgresql/tests/Flow/PostgreSql/Tests/Mother/MapperContextMother.php
```diff
@@ -0,0 +1,44 @@
...
+    public static function any() : Context
+    {
+        return new Context(
+            new Query('SELECT 1'),
+            new StubClient(),
+            new ClientContext(),
+        );
+    }
+
+    public static function with(
```
stloyd, line 21: Also by default `any()` and `with()` do the same.

src/lib/postgresql/src/Flow/PostgreSql/Client/Context.php
```diff
@@ -0,0 +1,78 @@
...
+    public function has(string $key) : bool
+    {
+        return \array_key_exists($key, $this->data);
+    }
+
+    public function merge(self $other) : self
+    {
+        return new self(
+            catalog: $other->catalog ?? $this->catalog,
+            data: \array_replace($this->data, $other->data),
```
stloyd, line 64: Sure it should do replace? Method says merge.

stloyd (approved):

src/lib/postgresql/src/Flow/PostgreSql/Client/Context.php
norberttech: yes, thats what array_replace does: 

> array_replace() creates a new array and assigns items into it for each key in each of the provided arrays. If a key appears in multiple input arrays, the value from the right-most input array will be used.

src/lib/postgresql/tests/Flow/PostgreSql/Tests/Mother/MapperContextMother.php
norberttech: it's cosmetic, irrelevant, if anything maybe rector could enforce the style here but in general there are places among the codebase that does the same

src/lib/postgresql/tests/Flow/PostgreSql/Tests/Mother/MapperContextMother.php
norberttech: yes and no, difference between `with` and `any` is that `any` should be used in the tests that arguments doesn't matter as they are not being used in the assertions. 
With on the other hand should be used when arguments are in some way relevant to the test, but since not all of them always are required, for simplicity with takes all arguments optional

PR #2408, opened 2026-05-31
Flow PHP - SEAL Adapter 

### LINKED ISSUE #2168: [Proposal]: Replace Meilisearch & Elasticsearch with SEAL
### Describe the Proposal

As we already spoke on Discord, I think we should remove Meilisearch and potentially Elasticsearch.

If we want to think about any future search engine, we should consider [SEAL](https://php-cmsig.github.io/search/) (I think we were speaking about it a long time ago as well...)

### API Adjustments

_No response_

Maybe

None

_No response_

MrHDOLEK:
The Elasticsearch and Meilisearch adapters are replaced by a single **SEAL** adapter
  (`flow-php/etl-adapter-seal`) built on top of [SEAL](https://github.com/php-cmsig/seal)
  — a PHP Search Engine Abstraction Layer. One adapter now covers Elasticsearch, OpenSearch,
  Meilisearch, Algolia, Solr, Typesense, RediSearch and Loupe: the user builds a
  `CmsIg\Seal\EngineInterface` and passes it to the DSL, exactly like the PostgreSQL/Doctrine
  adapters take a client. Tests are backend-agnostic (Memory adapter) and prove every Flow
  Entry type survives a round-trip through the adapter.
  
  Resolves: #2168
  
  
  <h2>Change Log</h2>
  <hr />
  <div id="change-log">
    <h4>Added</h4>
    <ul id="added">
      <li>SEAL adapter (<code>flow-php/etl-adapter-seal</code>) — a single, strongly typed integration for SEAL search engines (Elasticsearch, OpenSearch, Meilisearch, Algolia, Solr,
  Typesense, RediSearch, Loupe)</li>
      <li><code>from_seal()</code> extractor and <code>to_seal()</code> loader, working with any SEAL <code>EngineInterface</code></li>
      <li><code>to_seal_schema()</code> and <code>seal_schema_to_flow()</code> DSL — recursive, bi-directional Flow Schema ↔ SEAL Schema conversion (nested structures, lists and
  maps)</li>
      <li><code>seal_create_index()</code>, <code>seal_drop_index()</code>, <code>seal_create_schema()</code> and <code>seal_drop_schema()</code> DSL index-lifecycle helpers</li>
    </ul>
    <h4>Fixed</h4>
    <ul id="fixed">
    </ul>
    <h4>Changed</h4>
    <ul id="changed">
    </ul>
    <h4>Removed</h4>
    <ul id="removed">
      <li>Elasticsearch adapter (<code>flow-php/etl-adapter-elasticsearch</code>) — superseded by the SEAL adapter; use SEAL with the Elasticsearch backend
  (<code>cmsig/seal-elasticsearch-adapter</code>)</li>
    </ul>
    <h4>Deprecated</h4>
    <ul id="deprecated">
    </ul>
    <h4>Security</h4>
    <ul id="security">
    </ul>
  </div>


norberttech:
@MrHDOLEK as we discussed offline, please find some feedback and guidance below. 

# Extractor / Loader

In general I like the idea of having one extractor that accepts Seal EngineInterface which is aligned with how Doctrine Adapter works right now. Same goes for the loaders. 

It would also work very well with frameworks as seal provides a bundle for symfony that literally configures engine so to use it with flow it would be a simple matter of dependency injection. 

# What to test 

I had to think about it a bit more but I dont think we should retest every backend that seal is testing. Doctrine adapter is testing MySQL, PostgreSql nad Sqlite because flow-php/docrtine-dbal-bulk which extends doctrine default behaviors that needs to be tested. 

At this time seal provides following adapters: 

- [Elasticsearch](https://php-cmsig.github.io/search/research/index.html#elasticsearch) - cmsig/seal-elasticsearch-adapter
- [Opensearch](https://php-cmsig.github.io/search/research/index.html#opensearch) - cmsig/seal-opensearch-adapter
- [Meilisearch](https://php-cmsig.github.io/search/research/index.html#meilisearch) - cmsig/seal-meilisearch-adapter
- [Algolia](https://php-cmsig.github.io/search/research/index.html#algolia) - cmsig/seal-algolia-adapter
- [Solr](https://php-cmsig.github.io/search/research/index.html#solr) - cmsig/seal-solr-adapter
- [RediSearch](https://php-cmsig.github.io/search/research/index.html#redisearch) - cmsig/seal-redisearch-adapter
- [Typesense](https://php-cmsig.github.io/search/research/index.html#typesense) - cmsig/seal-typesense-adapter
- [Loupe](https://php-cmsig.github.io/search/research/index.html#loupe) - cmsig/seal-loupe-adapter

I would chose one of them (that requires the least configuration and is the easiest to setup in docker) and cover it with proper integration tests that would confirm that moving data from flow to seal works as expected.
Heck you can even use MemoryAdapter in tests unless other adapters have some features that requires some custom configuration. 
The only thing you might want to research first is are there any differences between seal adapters. Flow needs to be flexible enough to use all of them, so if one requires something custom, flow extractor/loader needs to be able to pass it through (but I dont think thats the case). 


# How to test 

So in general we want to cover in tests that seal adapter can handle all types of flow Entries. You can achieve that by using one of the two extractors: 

- [FakeStaticOrdersExtractor](https://github.com/flow-php/flow/blob/1efdd51abbb841068a56d98438d3bae048b92ba0/src/core/etl/tests/Flow/ETL/Tests/Double/FakeStaticOrdersExtractor.php)
- [FakeRandomOrdersExtractor](https://github.com/flow-php/flow/blob/aa83ce4cf5777e5e08a0cf6a1e6254f6f114b48e/src/core/etl/tests/Flow/ETL/Tests/Double/FakeRandomOrdersExtractor.php)

They both expose static method `schema() : Schema` but here is where the tricky part starts. I believe you need to be able to first create an index in tests. 
And that brings us to a missing piece of this PR - SchemaConverter.

## SchemaConverter 

Schema converters are recursive algorithms that can convert schema in both directiosn Flow to Seal and Seal to Flow. You can find some inspirations here: 

- [Parquet](https://github.com/flow-php/flow/blob/d3a74d087a7168cc53e80facae79f849bf9a2bef/src/adapter/etl-adapter-parquet/src/Flow/ETL/Adapter/Parquet/SchemaConverter.php)
- [Doctrine](https://github.com/flow-php/flow/blob/e466b8aa6a94611a46aafab91c3f4db19cc18c05/src/adapter/etl-adapter-doctrine/src/Flow/ETL/Adapter/Doctrine/SchemaConverter.php) 
- [PostgreSql](https://github.com/flow-php/flow/blob/a9c7b43267b9b9f685c02d269035ef60f9bd2429/src/adapter/etl-adapter-postgresql/src/Flow/ETL/Adapter/PostgreSql/SchemaConverter.php)

It might be the easiest to use LLM to help you create one based on those 3 examples for Seal (it's a recursive brain damaging exercise that might not be worth spending time on). 

Of course schema converters would need a DSL method. 

## Search Engine in Tests

So in this PR you are using traits to configure backends in tests, which is fine but there is a different pattern which I found cleaner and easier to maintain. 

Contexts. 

Here is a good example of [DatabaseContext](https://github.com/flow-php/flow/blob/e466b8aa6a94611a46aafab91c3f4db19cc18c05/src/cli/tests/Flow/CLI/Tests/Context/DatabaseContext.php) that is used in [DatabaseTableListCommandTest](https://github.com/flow-php/flow/blob/e466b8aa6a94611a46aafab91c3f4db19cc18c05/src/cli/tests/Flow/CLI/Tests/Integration/DatabaseTableListCommandTest.php) 

If there will be more than one integration tests you can extract an abstract `SealTestCase extends FlowTestCase` and setup SealContext in the `setUp` method making it available through `sealContext() : SealContext` method. 

In case of any questions, you know where to find me 😁

documentation/upgrading.md
```diff
@@ -799,9 +799,11 @@ typed(42, ValueType::INT4);
 
 ## Upgrading from 0.31.x to 0.32.x
 
-### 1) Removal of Meilisearch Adapter
+### 1) Removal of Meilisearch & Elasticsearch Adapters
```
stloyd: This must be a new note, that release was already done.

phpunit.xml.dist
```diff
@@ -23,7 +23,6 @@
         <env name="PGSQL_DATABASE_URL" value="pgsql://postgres:postgres@127.0.0.1:5432/postgres?serverVersion=11%26charset=utf8"/>
         <env name="MYSQL_DATABASE_URL" value="mysql://mysql:mysql@127.0.0.1:3306/mysql"/>
         <env name="SQLITE_DATABASE_PATH" value="./var/db/sqlite/test.db"/>
-        <env name="ELASTICSEARCH_URL" value="localhost:9200"/>
         <env name="MEILISEARCH_URL" value="http://localhost:7700"/>
         <env name="MEILISEARCH_API_KEY" value="masterKey"/>
```
stloyd: Can be removed as well.

src/adapter/etl-adapter-seal/.github/workflows/readonly.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  pull_request_target:
+    types: [opened]
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: superbrothers/close-pull-request@v3
```
stloyd: We now use directly GH CLI, Please see other adapters.

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealExtractor.php
```diff
@@ -0,0 +1,89 @@
...
+    public function extract(FlowContext $context): Generator
+    {
+        $offset = 0;
+
+        while (true) {
+            $search = $this->engine->createSearchBuilder($this->index);
+
+            if ($this->searchBuilder !== null) {
+                ($this->searchBuilder)($search);
+            }
```
stloyd: Should this be in loop? It means that whole search query is rebuilt on every page change

src/adapter/etl-adapter-seal/.github/workflows/readonly.yaml
MrHDOLEK: done

phpunit.xml.dist
MrHDOLEK: done

documentation/upgrading.md
MrHDOLEK: done

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealExtractor.php
MrHDOLEK: done

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/SealTestCase.php
```diff
@@ -0,0 +1,44 @@
...
+
+    private ?SealContext $sealContext = null;
+
+    protected function tearDown(): void
+    {
+        $this->sealContext?->dropIndex();
+        $this->sealContext = null;
+    }
+
+    protected function schema(): Schema
```
norberttech: I would avoid any kind of central schema at the test suit level. It's creating similar problem to tests that depends on db loaded fixtures. All tests using this are now fully depending on that schema which makes them fragile as any adjustment to this schama might break them. 

Each test should define its own schema and SealTestCase should only expose the Contexts, in this case SealContext

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/Context/SealContext.php
```diff
@@ -0,0 +1,50 @@
...
+        Schema $schema,
+        private readonly string $indexName,
+    ) {
+        $this->engine = new Engine(new MemoryAdapter(), $schema);
+
+        if ($this->engine->existIndex($this->indexName)) {
+            seal_drop_index($this->engine, $this->indexName);
+        }
+
+        seal_create_index($this->engine, $this->indexName);
```
norberttech: so in general I dont think that SealContext shold expect Schema, oposite, I think it should just expose methods that would simplify creating the engine for specific test. 

For example: 

SealContext::engine(Seal\Schema $schema)

and then in each test that requires a specific schema you should actually define a FlowSchema and use `to_seal_schema` to convert FlowSchema automatically to Seal Schema

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/Integration/SealAllEntryTypesTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $notes = $document['notes'];
+        static::assertCount(3, $notes);
+
+        /** @var list<array<string, mixed>> $items */
+        $items = $document['items'];
+        static::assertSame('SKU_0001', $items[0]['sku']);
+        static::assertSame(1, $items[0]['quantity']);
+    }
+
+    protected function schema(): Schema
```
norberttech: we should try to avoid at all cost protected/private methods in tests

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/Integration/SealAllEntryTypesTest.php
```diff
@@ -0,0 +1,47 @@
...
+use Flow\ETL\Adapter\Seal\Tests\SealTestCase;
+use Flow\ETL\Tests\Double\FakeStaticOrdersExtractor;
+
+use function Flow\ETL\Adapter\Seal\to_seal;
+use function Flow\ETL\Adapter\Seal\to_seal_schema;
+use function Flow\ETL\DSL\flow_context;
+
+final class SealAllEntryTypesTest extends SealTestCase
+{
+    public function test_round_trip_of_all_flow_entry_types(): void
```
norberttech: I would make this test heavier, I would build here a proper ETL pipelines that:

### Indexing Pipeline

1) extract from StaticOrdersExtractor (like 100 rows)
2) load to Seal 
3) Assert inserted rows (passing analyze() to makes DF return report)

### Extracting Pipeline 

1) Extract from seal 
2) fetchToArray 
3) assert

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/Unit/SchemaConverterTest.php
```diff
@@ -0,0 +1,122 @@
...
+use function Flow\Types\DSL\type_integer;
+use function Flow\Types\DSL\type_list;
+use function Flow\Types\DSL\type_string;
+use function Flow\Types\DSL\type_structure;
+
+final class SchemaConverterTest extends FlowTestCase
+{
+    public function test_converting_flow_schema_to_a_seal_index(): void
+    {
+        $sealSchema = (new SchemaConverter())->toSealSchema(
```
norberttech: in tests we should use DSL methods

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/RowsNormalizer.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Adapter\Seal;
+
+use Flow\ETL\Adapter\Seal\RowsNormalizer\EntryNormalizer;
+use Flow\ETL\Rows;
+use Generator;
+
+final readonly class RowsNormalizer
```
norberttech, line 11: this one is missing unit tests but even more important, EntryNormalizer is missing them

documentation/components/adapters/seal.md
```diff
@@ -0,0 +1,23 @@
...
+allowing you to extract from or load data into any Seal-supported search backend without coupling your pipelines to a
+specific search engine implementation.
+
+## Installation
+
+For detailed installation instructions, see the [installation page](/documentation/installation/packages/etl-adapter-seal.md).
+
+## Description
+
+ETL Adapter that provides Loaders and Extractors that work with any search engine supported by Seal.
```
norberttech: We gonna need more than this :D

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/functions.php
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[DocumentationDSL(module: Module::SEAL, type: Type::EXTRACTOR)]
+function from_seal(EngineInterface $engine, string $index): SealExtractor
+{
+    return new SealExtractor($engine, $index);
+}
+
+#[DocumentationDSL(module: Module::SEAL, type: Type::HELPER)]
+function seal_create_index(EngineInterface $engine, string $index): void
```
norberttech: I dont think we should be exposing those `seal_` functions, they are rather Seal related than Flow. DSL purpose is to cover Flow features

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealLoader.php
```diff
@@ -0,0 +1,54 @@
...
+            return;
+        }
+
+        $context->telemetry()->loadingStarted($this);
+
+        try {
+            $this->engine->bulk(
+                $this->index,
+                (new RowsNormalizer(new EntryNormalizer()))->normalize($rows),
+                [],
```
norberttech: this stands for deleteDocumentIdentifiers - similarly to how Docrtine and Flow PostgreSql is done, we should also allow to delete documents not only index them with this SealLoader

compose.yml.dist
```diff
@@ -26,18 +26,6 @@ services:
             MYSQL_ROOT_PASSWORD: root
         networks:
             - flow-php
-    elasticsearch:
```
norberttech, line 29: are we sure we dont wanna keep at least one real search engine in the repo? I mean I know we should be able to trust seal it works exactly same way but... Maybe keeping elasticsearch wouldnt be a bad idea and use it in integration tests keeping memory adapter for unit tests? (no strong opinion about this one yet, I never used seal so dont know how much in memory implementation is different from actual one)

norberttech (changes requested):
Looks much better! I left some comments, nothing critical, but there is one gap related to Delete operation. 

Please let me know if you have any questions about those comments!

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/Operation.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Adapter\Seal;
+
+enum Operation: string
+{
+    case DELETE = 'delete';
```
stloyd, line 9: It's more replace than delete cause data will be available, no?

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealExtractor.php
```diff
@@ -0,0 +1,91 @@
...
+                return;
+            }
+
+            $signal = yield array_to_rows($documents, $context->entryFactory());
+
+            if ($signal === Signal::STOP) {
+                return;
+            }
+
+            if (count($documents) < $this->pageSize) {
```
stloyd: Instead of counting, add counter variable and increment it in loop?

src/adapter/etl-adapter-seal/tests/Flow/ETL/Adapter/Seal/Tests/Context/SealContext.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        $this->indexes = [];
+    }
+
+    public function engine(Schema $schema): EngineInterface
+    {
+        $engine = new Engine(new MemoryAdapter(), $schema);
+
+        foreach (array_keys($schema->indexes) as $index) {
```
stloyd:
```suggestion
        foreach ($schema->indexes as $index => $value) {
```

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/Operation.php
norberttech: no, it's actual delete

norberttech (approved):
Looks awesome @MrHDOLEK ! 

If you feel that work is completed, and this adapter is now usable I'm happy to merge it 🚀

MrHDOLEK:
@norberttech I found an issue while testing with the 20k dataset.

SEAL cannot read data beyond 10K from elastic because it only supports offset-based pagination, and Elasticsearch blocks offsets outside the 10K result window. To work around this problem, I added cursor-based pagination.

norberttech:
> SEAL cannot read data beyond 10K from elastic because it only supports offset-based pagination, and Elasticsearch blocks offsets outside the 10K result window. To work around this problem, I added cursor-based pagination.

That's excellent finding! 

Does other engines also comes with such limitations? Is this keyset pagination going to work only with Elastic or all seal adapters? 
When someone tries to read above 10k values without keyset, is the exception error going to point him to the keyset? Are docs updated? 
What is the cost of moving to such pagination? Is it really keyset or maybe a cursor pagination? Are there other types of pagination supported? 

@alexander-schranz if by any chance you could take a look at this integration, your feedback would be greatly appreciated 🙏

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealExtractor.php
```diff
@@ -0,0 +1,200 @@
...
+        $cursor = null;
+
+        while (true) {
+            $search = $this->engine->createSearchBuilder($this->index);
+
+            if ($this->searchBuilder !== null) {
+                ($this->searchBuilder)($search);
+            }
+
+            $search->addSortBy($cursorField, $direction->value)->limit($this->pageSize)->offset(0);
```
stloyd: Should be out of the loop

src/adapter/etl-adapter-seal/src/Flow/ETL/Adapter/Seal/SealExtractor.php
```diff
@@ -0,0 +1,200 @@
...
+                        : Condition::greaterThan($cursorField, $cursor),
+                );
+            }
+
+            $documents = [];
+            $count = 0;
+
+            foreach ($search->getResult() as $document) {
+                $documents[] = $document;
+                $cursor = $this->cursorValue($document, $cursorField);
```
stloyd: Shouldn't this value be only check when count is equal page size? Cause otherwise it will not be used anyway yet it will check 999 times value with page limit of 1k and only 999 results, no?

MrHDOLEK:
> > SEAL cannot read data beyond 10K from elastic because it only supports offset-based pagination, and Elasticsearch blocks offsets outside the 10K result window. To work around this problem, I added cursor-based pagination.
> 
> That's excellent finding!
> 
> Does other engines also comes with such limitations? Is this keyset pagination going to work only with Elastic or all seal adapters? When someone tries to read above 10k values without keyset, is the exception error going to point him to the keyset? Are docs updated? What is the cost of moving to such pagination? Is it really keyset or maybe a cursor pagination? Are there other types of pagination supported?
> 
> @alexander-schranz if by any chance you could take a look at this integration, your feedback would be greatly appreciated 🙏
1.
Algolia — 1 000 (paginationLimitedTo, max 20k)
Meilisearch — 1 000 (maxTotalHits)
Typesense — max_per_page 250 + limit_hits ~500
2. A raw Elasticsearch error is returned. There is currently no error handling (catch) for this exception on the SEAL side.
3. I would not worry about performance here. The real problem is its pagination limitations.
4. This could be solved by adding some primitive scroller in SEAL.

norberttech:
> 4. This could be solved by adding some primitive scroller in SEAL.

In general pagination is a complex problem, especially over large datasets.
There are 3 patterns I'm using in PostgreSql and Doctrine adapters. 

Limit Offset Pagination - usually works up to 5-10k rows 
Cursor Pagination - you need to understand cursors first as they might expire while you are paginating and they might highly impact the performance (cursors should be configurable) 
KeySet Pagination - my usual default but storage must support it. 

I guess the biggest question is if SEAL would let us now build something that would leverage KeySet pagination without any changes in SEAL or maybe some adjustments are first required there?

alexander-schranz:
@norberttech I can just give a quick response here without having a deeper look at the Pull Request yet. Limitations on Elasticsearch can be controlled by Elasticsearch settings, so if somebody want to fetch more they can increase it on Elasticsearch instance config. There are some discussions to allow configure more settings via DSN params, but that will take a longer time.

While there is an issue around adding a `Scroller`: https://github.com/PHP-CMSIG/search/issues/535 it won't solve the Elasticsearch 10k issue, as @MrHDOLEK mention the offset is blocked by Elastic and the Scroller issue is designed that it not need be implemented per Adapter = uses Limit / Offset like Scrolling. Maybe something we need to rethink.

> KeySet Pagination

Are you in control of the Schema and the Data? If I understand the KeySet Pagination correctly you would require additional field or a value which increases. If document id is integer maybe sort by id, else a datetime field or something which can be sorted by and is possible unique or atleast not exist more as 250 (typesense default limit).

@norberttech How are you handling KeySet Pagination in doctrine if the identifier are not auto incremented like uuids?

norberttech:
> @norberttech I can just give a quick response here without having a deeper look at the Pull Request yet. 

Thank you! 

> Limitations on Elasticsearch can be controlled by Elasticsearch settings,

That's unfortunately not acceptable solution in data processing world, it does not scale and it's also not very complex to hit that limit even when the scale is not massive + there is a limit of how much you can increase those limits up to. In fact, 10k rows is actually a micro dataset, even increasing to 100k - this is nothing. 

> @norberttech How are you handling KeySet Pagination in doctrine if the identifier are not auto incremented like uuids?

Yes, KeySet pagination requires something you can sort by, it can be either auto-incremented id, Uuid V7, additional timestamp column.
I don't think I ever had to deal with a dataset that wouldn't have at least one of those, and even if not, adding something as small as `createdAt` timestamp solves the problem usually. 

> Are you in control of the Schema and the Data?

Of course, if you don't have a control over it, you can't build an reliable ETL pipeline.

norberttech:
I took a walk and tried to think about it from a different perspective :D 

Instead of trying to solve pagination problem, maybe we should abandon the idea of having SEALExtractor in the first place? 

ETL - stands for Extract / Transform / Load - but what use case one could have to Extract entire index? 

SearchEngines are rarely a source where data lives (and if they do, it's a bigger architectural problem), so streaming data from them (as this is what we Extractors are doing). 

So how about we keep SealLoader which is anyway has only real use case and strong justification I can think about now, drop the idea of SealExtractor and only tackle this once we get a real and solid use case for it? 

@stloyd @MrHDOLEK - what do you guys think about it?

alexander-schranz:
> Yes, KeySet pagination requires something you can sort by, it can be either auto-incremented id, Uuid V7, additional timestamp column.

Do you configure in Flow Doctrine which column is used then for the KeySet pagination? If you do that, you can do the same for the SEAL Extractor it would work atleast for integer fields and datetime fields. For string fields the support is differently Meilisearch, Algolia has no support for lexicalgraphical string comparison `> 'ABC'` so uuid would not work.

So code for such case would look like:

```php
$indexName = 'blog'; // From Flow Config
$keySetPaginationField = 'createdAt'; // From Flow Config
$keySetValue = 0; // From Flow Config; ?

do {
    $result = $this->engine->createSearchBuilder($indexName)
        ->addFilter(Condition::greaterThan($keySetPaginationField, $keySetValue))
        ->addSortBy($keySetPaginationField, 'asc')
        ->setLimit(100)
        ->getResult();
    
    foreach ($result as $document) {
        $keySetValue = $document[$keySetValue]
    
        yield $document;
    }
} while (\count($result) > 0);
```

This way you should never hit the limits by the search engines.

norberttech:
> So code for such case would look like:
> 
> ```
> $indexName = 'blog'; // From Flow Config
> $keySetPaginationField = 'createdAt'; // From Flow Config
> $keySetValue = 0; // From Flow Config; ?
> 
> do {
>     $result = $this->engine->createSearchBuilder($indexName)
>         ->addFilter(Condition::greaterThan($keySetPaginationField, $keySetValue))
>         ->addSortBy($keySetPaginationField, 'asc')
>         ->setLimit(100)
>         ->getResult();
>     
>     foreach ($result as $document) {
>         $keySetValue = $document[$keySetValue]
>     
>         yield $document;
>     }
> } while (\count($result) > 0);
> ```
> 
> This way you should never hit the limits by the search engines.

Yes, that should work just fine! In postgresql/doctrine we do something like this: 

```php
<?php

df()
    ->read(from_pgsql_key_set(
        $client,
        "SELECT id, name, email FROM users",
        pgsql_pagination_key_set(pgsql_pagination_key_asc('id')),
        pageSize: 1000
    ))
    ->write([to_output](https://flow-php.com/documentation/dsl/core/to-output/#dsl-function)())
    ->run();
```

So the extractor expects a pagination keyset `pgsql_pagination_key_set()` which expects a list of columns with order `pgsql_pagination_key_asc('id')` 

But after some internal conversations we decided to put a side for now Extractor until someone will come with a legit use case for it. This should reduce also a maintenance and code volume of this adapter. 

Thanks for your input @alexander-schranz !

alexander-schranz:
@norberttech why it makes me a little bit sad, I totally understand that decision if there is no need currently for Search Index based Adapter in Flow.

If you need something always feel free to ping me. I added the findings of this discussion to the `Scroller` issue of SEAL: https://github.com/PHP-CMSIG/search/issues/535#issuecomment-4727842639 and will sure take that in mind when implementing such thing.

If there is any other feedback you or the others have around SEAL it is always welcome.

norberttech:
> @norberttech why it makes me a little bit sad, I totally understand that decision if there is no need currently for Search Index based Adapter in Flow.

Oh we are not dropping the adapter, it's extremaly value and I wanted to have it since I found your project, we are just reducing the scope of this pr slightly by dropping Extraction part of the (ETL) - Transform and Load stays! 

If you have a legit use cases for traversing over search engine results in ETL pipeline, we are more than happy to implement that extractor with the keyset pagination embedded, just say a word 😊

This whole adapter even without Extractor will bring a massive value, I just feel like at this point Extractor might increase the complexity of the adapter, that might not be worth the effort without a valid use case first. 

SealLoader is unquestionable valuable addition to the framework 💪 Thank you for building SEAL!

alexander-schranz:
> Oh we are not dropping the adapter, it's extremaly value and I wanted to have it since I found your project, we are just reducing the scope of this pr slightly by dropping Extraction part of the (ETL) - Transform and Load stays!

I see that make sense. 😃 

> If you have a legit use cases for traversing over search engine results in ETL pipeline, we are more than happy to implement that extractor with the keyset pagination embedded, just say a word 😊

I will! If there are any changes in SEAL around the paginiation I will keep you uptodate when we add KeySet Pagination / Scroller to SEAL Core.

> SealLoader is unquestionable valuable addition to the framework 💪 Thank you for building SEAL!

Nice to hear, looking forward to it. Let me know when it is released, I'm thinking to add a Section to the SEAL docs mentioning the SEAL Adapter in Flow PHP ETL.

norberttech:
@alexander-schranz the PR was merged, SEAL Adapter is now available as `1.x-dev` and will become available under `0.41.0` release in next 2 weeks! 

Documentation is available here: https://flow-php.com/documentation/components/adapters/seal/ 
Packagist: https://packagist.org/packages/flow-php/etl-adapter-seal

I'm going to spread the word at my social media in a moment!

PR #2409, opened 2026-05-31
Add git detector to telemetry

jdecool:
<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add git detector to telemetry</li>
    <li>Add git detector to phpunit telemetry</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

<h2>Description</h2>
<hr>
Sending PHPUnit data is very useful to see what happens during test execution.

But when used in an active project, where there are a lot of tests running, it's very interesting to have Git repository-related information.

So this PR adds a Git detector to attach some VCS information. As it could be interesting outside of the PHPUnit context, I've added it to the telemetry module.


src/lib/telemetry/tests/Flow/Telemetry/Tests/Integration/Resource/Detector/GitRepositoryFactory.php
```diff
@@ -0,0 +1,81 @@
...
+
+        if ($branch !== null) {
+            @shell_exec($git . ' branch -c ' . escapeshellarg($branch) . ' 2>/dev/null');
+            @shell_exec($git . ' checkout -q ' . escapeshellarg($branch) . ' 2>/dev/null');
+        }
+
+        if ($tag !== null) {
+            @shell_exec($git . ' tag ' . escapeshellarg($tag) . ' 2>/dev/null');
+            @shell_exec($git . ' checkout -q ' . escapeshellarg($tag) . ' 2>/dev/null');
+        }
```
norberttech: I'm honestly not very comfortable with all those shell execs in the test suite. 
I'm also tryign to use local `__DIR__ . /var` folders close to tests rather than `sys_get_temp_dir`. 

Personally I think it would be better if those git repositories would live in some `Fixtures` folder in the integration test suite instead of creating them on demand

src/lib/telemetry/src/Flow/Telemetry/Resource/Detector/GitDetector.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $output = @shell_exec($command);
+
+        if (!is_string($output)) {
+            return null;
+        }
+
+        $output = trim($output);
+
+        return $output === '' ? null : $output;
```
norberttech: let me think about this one, it might expose some security details. 
For example here: `$repositoryUrl = $this->runGit('git remote get-url origin');` 
If one is keeping it like this `https://user:password@host/repo.git`. 

It also seems that `$this->runGit('git remote get-url origin');` would create `git git remote get-url origin` command. 

Another issue I see here is that `git` is resolved from `$PATH` which might not be the always the case so users should be able to pass also additional git binary path. 

I'm also not sure that `shell_exec` is the best option here, I would probably try to do it with proc_open as it does not need to even go through shell 

```php
private function runGit(array $args): ?string
{
    $cmd = ['git'];
    if ($this->workingDirectory !== null) {
        $cmd[] = '-C';
        $cmd[] = $this->workingDirectory;
    }
    $cmd = [...$cmd, ...$args];
    // run with proc_open / Process, no shell
}
```

norberttech:
@jdecool I really appreciate the PR but it's a bigger addition that should go through a [Proposal Process](https://github.com/flow-php/flow/issues/new?template=proposal.yml) since there are many architectural choices required in the implementation (like how to call git for example). 
There is also a part that is not covered at all, it should be also possible to configure this detector through symfony telemetry bundle (which could be another task and this is what the proposal process is for).

We can keep this PR and work on it directly, but in the future I would appreciate opening a proper Proposal so we can first talk about how to implement the feature before writing any code

jdecool:
Thanks for your review and your comments.

No problem, I'm going to start the proposal process.

To explain this PR, I've initially planned to try to add custom detectors dynamically during the PHPUnit process, but I think that git detector could be added by default.

But you are right, it's a big change and should be discussed before implementing it.

norberttech:
> To explain this PR, I've initially planned to try to add custom detectors dynamically during the PHPUnit process, but I think that git detector could be added by default.

I fully agree, git detector should be part of the telemetry, it should be optional though, but also exposed to the whole telemetry not only phpunit bridge 

> But you are right, it's a big change and should be discussed before implementing it.

I'm glad we are on the same page!

norberttech:
@jdecool if you would like to brainstorm this first, or have any questions about implementation, scope etc, you can always find me on Flow discord: https://discord.gg/flow-php or you can schedule with me a [session online](https://calendar.proton.me/bookings#EyjcEe07NRPoD6ECOGt-Qd--PRYh7XmulLqtf5l2zKE=)

jdecool:
Thanks @norberttech I've joined the discord.

I plan to work on this PR and the proposal process this weekend.

src/bridge/phpunit/telemetry/src/Flow/Bridge/PHPUnit/Telemetry/TelemetryFactory.php
```diff
@@ -40,6 +41,7 @@ public static function create(Configuration $config): Telemetry
     {
         $telemetryResource = resource_detector()
```
norberttech, line 42: I would pass `git_detector()` here  with other default detectors: 
```
resource_detector([
  os_detector(),
  host_detector(),
  process_detector(),
  composer_detector(),
  environment_detector(),
  git_detector() 
]);
```

but ideally (can be handled in next pr) we should probably  allow users to configure which detectors they want to use

src/lib/telemetry/src/Flow/Telemetry/Resource/Detector/GitDetector.php
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Removes any embedded credentials (user:password@) from a remote URL so they
+     * are never reported as a resource attribute.
+     *
+     * SCP-like SSH remotes (e.g. "git@github.com:org/repo.git") carry no secret and
+     * are returned untouched, as are URLs that cannot be parsed.
+     */
+    private function sanitizeRepositoryUrl(string $url): string
```
norberttech: I'm not sure how much I like that sanitization is embedded into GitDetector which seems to be breaking his single responsibility. 

I think this whole sanitization logic should be extracted to standalone class, properly unit tested and used in GitDetector as `(new RemoteUrlSanitizer())->sanitize($remoteUrl)`

norberttech:
its looking good @jdecool !

jdecool:
Sorry, I haven't been around much lately; I was really busy last week.

I'll finish this pull request soon. Thanks for your feedback and help.

norberttech:
> Sorry, I haven't been around much lately; I was really busy last week.
> 
> I'll finish this pull request soon. Thanks for your feedback and help.

Take your time, it's a marathon, not a sprint. I always prefer to wait than merge something not fully baked. Thank you!

jdecool:
Thank you for your patience @norberttech .

I've made some suggested changes; please take a look and let me know if it's OK for you.

norberttech:
> Thank you for your patience @norberttech .
> 
> I've made some suggested changes; please take a look and let me know if it's OK for you.

Thanks @jdecool! 
There are few small things that should be fixed, but I'm going to take care of this. 
I'm going to merge it tomorrow morning and right after that will push those minor improvements (like we dont really user Helpers in tests, we create Contexts)

jdecool:
Thanks @norberttech 

You're doing an amazing job on this project ! 🙏

PR #2423, opened 2026-06-04
chore: Move Telemetry `SerializerType` enum from PHPUnit to OTLP

stloyd:
Serializers are defined in OTLP, not in the PHPUnit bridge; thus, the enum with its definition should be in OTLP.

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Move Telemetry `SerializerType` enum from PHPUnit to OTLP</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


norberttech:
that enum exists only to validate serialization type in the bridge that exposes it through phpunit.xml configuration, whats the use case for keeping it in the OTLP bridge?

stloyd:
I'm writing PoC telemetry for Behat, and I found it useful to have available serializers in one place, and noticed that there is one, but in the PHPUnit package, but serializers are in OTLP in fact.

norberttech:
ah, then it makes sense!

norberttech:
failing tests are unrelated, I'm investigating this now

PR #2449, opened 2026-06-14
chore: Migrate Symfony XML configuration usage to the PHP one

stloyd:
Symfony [deprecated the XML configuration format](https://symfony.com/blog/new-in-symfony-7-4-deprecated-xml-configuration) in release 7.4, and removed it in 8.0. The recommended format is PHP one.

Should fix issues in tests like:
```
Flow\Bridge\Symfony\PostgreSqlBundle\Tests\Integration\Profiler\FlowPostgreSqlProfilerTest::test_panel_renders_executed_queries_in_the_profiler
Symfony\Component\Config\Exception\LoaderLoadException: Cannot load resource "@WebProfilerBundle/Resources/config/routing/profiler.xml". Make sure the "WebProfilerBundle" bundle is correctly registered and loaded in the application kernel class. If the bundle is registered, make sure the bundle path "@WebProfilerBundle/Resources/config/routing/profiler.xml" is not empty.
```

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Migrate Symfony XML configuration usage to the PHP one</li>
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>

### DIFF
--- a/src/bridge/symfony/postgresql-bundle/composer.json
+++ b/src/bridge/symfony/postgresql-bundle/composer.json
@@ -21,6 +21,8 @@
         "twig/twig": "^3.0"
     },
     "require-dev": {
+        "composer-runtime-api": "^2.2",
+        "composer/semver": "^3.4",
         "flow-php/phpunit-postgresql-bridge": "self.version",
         "flow-php/symfony-postgresql-cache-bridge": "self.version",
         "flow-php/symfony-postgresql-messenger-bridge": "self.version",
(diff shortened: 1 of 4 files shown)


norberttech:
Does it work also with 6.4? We are supporting right now 6.4 | 7.4 and 8.0+

stloyd:
Seems like the new type was added in 7.4, in 6.4 it's still XML one.

norberttech:
hmm let me think about it, we might be able to drop 6.4 entirely

PR #2473, opened 2026-06-21
#1390 -  configurable backend for group by buckets

### LINKED ISSUE #1390: Storage for Aggregating Functions
Currently, all Aggregating Functions are storing aggregated results in memory. 
This is not fully aligned with Flow core philosophy which says that Flow is supposed to be memory efficient in the first place. 

To solve that problem we should introduce a dedicated storage for mixed results of aggregating functions into which they can easily read/write to. 

It should be configurable, meaning that end user should be able to chose if he wants to aggregate values in memory (super fast but not scalable) or to use one of the available storage strategies. 

We can approach it in the similar way to ~GroupBy~ Sort: 

- when memory consumption < X - use memory 
- when memory consumption > X - use cache 

We might be better using dedicated storage rather than Cache which is optimized for Rows. In this case, Redis-based storage might be a very good solution.

### LINKED ISSUE #2472: [Proposal]: Bounded-memory GroupBy aggregation (streaming merge + streaming result)
## Describe the Proposal

Aggregation is a blocking stage, so GroupBy must hold per-group state until all input is consumed. Today that footprint is unbounded in two places:

1. `GroupBy::result()` materializes the **entire** result as a single `Rows` (one `Row`/`Entry`/`Definition`/`Type` graph per distinct group).
2. Even with the configurable spill storage from #1390, the `MEMORY_FALLBACK_EXTERNAL` backend does **not** bound peak memory: `ExternalAggregationStorage::all()` reloads every spilled bucket into one in-memory `$merged` map, and `result()` then builds a second full `Rows`. At high cardinality the "spill" backend therefore uses **more** peak memory than the in-memory one (measured with `groupBy('email')->aggregate(count(), sum())`: filesystem ≈ 432 MB vs memory ≈ 305 MB at ~98k groups; ~1.1 GB at 300k). Spilling bounds ingestion, then `all()`/`result()` un-bound it.

**Goal:** make GroupBy aggregation bounded end-to-end on the external path — `O(batchSize + bucketCount)` instead of `O(distinctGroups)` — and drop the redundant full-result copy for every backend. This is what makes `MEMORY_FALLBACK_EXTERNAL` actually deliver scaling beyond RAM. It also aligns GroupBy with Flow's streaming contract: it is currently the only operator that emits one giant batch instead of a stream of `Rows`.

Two coupled changes — **neither alone fixes the filesystem path** (one removes `$merged`, the other removes the result `Rows`; both are required):

- **Streaming k-way merge (storage):** spill buckets sorted by group hash (one record per entry), then merge them with a min-heap, yielding one fully-merged group at a time — no `$merged` map.
- **Streaming result (core):** `GroupBy` emits the result in batches instead of one `Rows`; the downstream pipeline already consumes per-`Rows` incrementally, so this flows through to file/stream loaders unchanged.

## API Adjustments

**No user-facing DataFrame/DSL API change** — internal behavior only; existing `groupBy()` / `aggregate()` usage is unchanged.

Internal:

- `GroupBucketsCache`: `get(): array` → `stream(string): Generator<string, GroupEntry>`; spill format becomes one sorted record per entry.
- `ExternalAggregationStorage::all()`: streaming k-way merge (reuses the `SplMinHeap` pattern from `ExternalSort`).
- `GroupBy`: add `resultStream(FlowContext, int $batchSize): Generator<Rows>`; `GroupByProcessor` does `yield from`. `result()` kept for the pivot path / `fetch()` / BC. Batch size from existing config.

Semantics decisions needed (the reason this touches core):

- **Output ordering:** the memory backend yields groups in insertion order; the external merge yields them sorted by group hash. GroupBy does not currently guarantee an order — needs a contract decision.
- **`to_output` table mode:** `StreamLoader` closes the output stream after the first batch and `AsciiTableFormatter` prints a header per batch, so multi-batch groupBy output breaks interactive table rendering. Requires fixing `StreamLoader` (write the header once, don't close mid-stream) or special-casing. Main cross-cutting change.
- `fetch()` unaffected (materializes by design); pivot stays materialized (small, separate path).

## Are you intending to also work on proposed change?

Yes

## Are you interested in sponsoring this change?

No

## Integration & Dependencies

No new dependencies. Reuses existing primitives (`SplMinHeap` as in `ExternalSort`, the configured `Serializer`, `Filesystem`).

MrHDOLEK:
Makes GroupBy aggregation memory-efficient and configurable (issue #1390): keep the fast in-memory path by default, or opt into an external, bounded-memory path.

Resolves: #1390
Resolves: #2472 because this pr resolve this problem.

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Configurable, memory-efficient GroupBy aggregation: <code>ConfigBuilder::aggregationAlgorithm()</code> and the <code>AggregationAlgorithms</code> enum select between <code>MEMORY_AGGREGATION</code> (default, in-memory hash map) and <code>EXTERNAL_AGGREGATION</code> (sorts by the group-by columns and folds, spilling to disk for bounded memory), with <code>aggregationFilesystem()</code> for the spill protocol.</li>
    <li>GroupBy streams its result in batches instead of materializing every group into a single <code>Rows</code>.</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/core/etl/src/Flow/ETL/GroupBy/Storage/GroupEntry.php
```diff
@@ -0,0 +1,19 @@
...
+
+final readonly class GroupEntry
+{
+    /**
+     * @param array<string, mixed> $values
+     * @param array<AggregatingFunction> $aggregators
+     */
+    public function __construct(
+        public array $values,
+        public array $aggregators,
```
norberttech: I'm not really sure if I understand this data structure, it's keeping values for all aggregators?

src/core/etl/src/Flow/ETL/Config/ConfigBuilder.php
```diff
@@ -79,6 +85,34 @@ public function __construct()
...
+    }
+
+    public function aggregationMemoryLimit(Unit $unit): self
+    {
+        $this->aggregation->memoryLimit($unit);
+
+        return $this;
+    }
+
+    public function aggregationStorage(AggregationStorageStrategy $strategy): self
```
norberttech: so storage or strategy? 😅

src/core/etl/src/Flow/ETL/Config/ConfigBuilder.php
```diff
@@ -79,6 +85,34 @@ public function __construct()
...
+    {
+        $this->aggregation->strategy($strategy);
+
+        return $this;
+    }
+
+    public function aggregationStore(AggregationStorage $storage): self
```
norberttech: Those two methods, aggregationStorage and aggregationStore are a bit confusing

src/core/etl/src/Flow/ETL/GroupBy/Storage/AggregationStorage.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * @param array<string, mixed> $values
+     * @param array<AggregatingFunction> $prototypes
+     */
+    public function entry(string $hash, array $values, array $prototypes): GroupEntry;
+
+    public function flush(): void;
+
+    public function save(string $hash, GroupEntry $entry): void;
```
norberttech: what's the difference between entry and save?

src/core/etl/src/Flow/ETL/Function/MergeableAggregatingFunction.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Flow\ETL\Function;
+
+interface MergeableAggregatingFunction extends AggregatingFunction
```
norberttech: could you walk me through this logic? Like when two AggregatingFunctions can be merged?

src/core/etl/src/Flow/ETL/GroupBy/HashAggregation.php
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * @param array<string, GroupState> $hot
+     *
+     * @return Generator<int, array{0: string, 1: GroupState}>
+     */
+    private function hotCursor(array $hot): Generator
+    {
+        foreach ($hot as $hash => $state) {
+            yield [$hash, $state];
```
norberttech: please don't use tuples for return types, introduce a simple VO instead

src/core/etl/src/Flow/ETL/Config/Aggregation/AggregationConfigBuilder.php
```diff
@@ -0,0 +1,44 @@
...
+        if ($this->memoryLimit === null) {
+            $envLimit = getenv(AggregationConfig::AGGREGATION_MAX_MEMORY_ENV);
+
+            $this->memoryLimit = is_string($envLimit) ? Unit::fromString($envLimit) : Unit::fromBytes(PHP_INT_MAX);
+        }
+
+        return new AggregationConfig($this->memoryLimit, $this->filesystemProtocol);
+    }
+
+    public function filesystemProtocol(string $protocol): self
```
norberttech: we are not really using `protocol`, instead I started using `mount` as there can be multiple filesystems that share protocol but they might be registered under different mount names. For example when you want to move S3 to S3 through FSTab

src/core/etl/src/Flow/ETL/Processor/GroupByProcessor.php
```diff
@@ -22,10 +24,24 @@ public function __construct(
...
 
-        yield $this->groupBy->result($context);
+        $config = $context->config->aggregation;
+
+        $aggregation = new HashAggregation(
+            $config->memoryLimit,
+            new FilesystemGroupBucketsCache(
+                $context->filesystem($config->filesystemProtocol),
+                $context->config->serializer(),
+                100,
```
norberttech: this should be configurable

src/core/etl/src/Flow/ETL/DataFrame/GroupedDataFrame.php
```diff
@@ -21,15 +21,15 @@ public function aggregate(AggregatingFunction ...$aggregations): DataFrame
...
-        // @mago-ignore analysis:invalid-method-access
-        $pipelineAdder->bindTo($this->df, $this->df)($this->groupBy);
+        return $this->addProcessor(new GroupByProcessor($this->groupBy));
+    }
 
-        return $this->df;
+    /**
+     * @param null|int<1, max> $size
+     */
+    public function getGroups(?int $size = null): DataFrame
```
norberttech: this should be removed, it does not belong to the GroupedDataFrame API

src/core/etl/src/Flow/ETL/Sort/ExternalSortFactory.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Flow\ETL\FlowContext;
+use Flow\ETL\Sort\ExternalSort\BucketsCache\FilesystemBucketsCache;
+use Flow\Filesystem\Filesystem;
+use Flow\Filesystem\Path;
+
+/**
+ * @internal
+ */
+final class ExternalSortFactory
```
norberttech: I don't think we need this

PR #2516, opened 2026-07-08
Add a new `ref()->domElementNamespace()` function

### LINKED ISSUE #2514: [Proposal]: New function for XML: DOM element namespace
### Describe the Proposal

When working with XML documents, you often need to know the namespace of particular elements to determine which type of validation should be applied or how to read the document. Currently, there is no way to get a namespace from XML nodes in Flow.

I would propose adding a new function similar to `DOMElementAttributeValue`, i.e., `DOMElementNamespaceValue`.

As an example, I would suggest, i.e., [UBL (Universal Business Language)](https://en.wikipedia.org/wiki/Universal_Business_Language): https://docs.oasis-open.org/ubl/os-UBL-2.1/UBL-2.1.pdf (page 2 with list of known schemas that determine document logic)

### API Adjustments

PoC (not tested at all)

```php
<?php

declare(strict_types=1);

namespace Flow\ETL\Function;

final class DOMElementNamespaceValue extends ScalarFunctionChain
{
    public function __construct(
        private readonly ScalarFunction|DOMNode|HTMLElement $domElement,
        private readonly ScalarFunction|string $attribute,
    ) {}

    public function eval(Row $row, FlowContext $context): ?string
    {
        $types = [
            type_instance_of(DOMNode::class),
            type_list(type_instance_of(DOMNode::class)),
        ];

        if (class_exists('\Dom\HTMLElement')) {
            $types[] = type_instance_of(HTMLElement::class);
            $types[] = type_list(type_instance_of(HTMLElement::class));
        }

        $node = (new Parameter($this->domElement))->as($row, $context, ...$types);

        if ($node instanceof DOMDocument) {
            $node = $node->documentElement;
        }

        if (is_array($node) && count($node)) {
            $node = reset($node);
        }

        if ($node === null) {
            return $context
                ->functions()
                ->invalidResult(new InvalidArgumentException('DOMElementNamespaceValue requires non-null DOMNode'));
        }

        $attributeName = (new Parameter($this->attribute))->asString($row, $context);

        if ($attributeName === null) {
            return $context
                ->functions()
                ->invalidResult(
                    new InvalidArgumentException('DOMElementNamespaceValue requires non-null attribute name'),
                );
        }

        if (!$node instanceof DOMNode && !$node instanceof HTMLElement) {
            return null;
        }

        $attributes = $node->attributes;

        if ($attributes === null) {
            return null;
        }

        foreach ($attributes as $attribute) {
            if ($attribute->nodeName === $attributeName || str_starts_with($attribute->nodeName, $attributeName.':')) {
                return $attribute->nodeValue;
            }
        }

        return null;
    }
}
```

Yes

No

None

stloyd:
Resolves: #2514

<h2>Change Log</h2>
<hr /> 
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
    <li>Add a new `ref()->domElementNamespace()` function</li>
    <li>Add support for a new `\Dom\XmlDocument` (PHP 8.4+) in XML handling</li>
  </ul> 
  <h4>Fixed</h4>  
  <ul id="fixed">
    
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    
  </ul>  
  <h4>Removed</h4>
  <ul id="removed">
    
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
    
  </ul>  
  <h4>Security</h4> 
  <ul id="security">
    
  </ul>     
</div>


src/core/etl/src/Flow/ETL/Function/DOMElementNamespaceValue.php
```diff
@@ -0,0 +1,90 @@
...
+            // @mago-ignore analysis:possibly-invalid-argument
+            $xpath = new XPath($node->ownerDocument);
+        } else {
+            // @mago-ignore analysis:possibly-invalid-argument
+            $xpath = new DOMXPath($node->ownerDocument);
+        }
+
+        /** @var \DOMNameSpaceNode $nsNode */
+        // @mago-ignore analysis:invalid-iterator
+        foreach ($xpath->query('namespace::*') ?: [] as $nsNode) {
```
stloyd, line 80: New version always returns a traversable object, old returned array or `false`, and Mago has a problem detecting which is correct.

src/core/etl/src/Flow/ETL/Row/Entry/XMLEntry.php
```diff
@@ -153,7 +154,12 @@ public function toString(): string
             return '';
         }
 
-        $serialized = $this->value->saveXML($this->value->documentElement);
+        if ($this->value instanceof XMLDocument) {
+            $serialized = $this->value->saveXml($this->value->documentElement);
+        } else {
+            // @mago-ignore analysis:possibly-invalid-argument,possibly-invalid-argument
+            $serialized = $this->value->saveXML($this->value->documentElement);
```
stloyd, line 161: Mago has a problem holding the Dom object type, even with an instance of pointing to its new or old version.

src/core/etl/src/Flow/ETL/Row/Entry/XMLElementEntry.php
```diff
@@ -168,7 +170,14 @@ public function toString(): string
             return '';
         }
 
-        $serialized = $ownerDocument->saveXML($this->value);
+        if ($ownerDocument instanceof XMLDocument) {
+            // @mago-ignore analysis:possibly-invalid-argument
+            $serialized = $ownerDocument->saveXml($this->value);
```
stloyd, line 175: In official docs, this method is not documented, but it's visible in [the official stub](https://github.com/php/php-src/blob/master/ext/dom/php_dom.stub.php#L1724).

PR #2524, opened 2026-07-11
refactor: replace switch statements with match expressions

MrHDOLEK:
## Summary

Replaces `switch` statements with `match` expressions.

## Changes

- `Flow\ETL\Dataset\Memory\Unit::fromString()` — dispatch and the default *digit-or-throw* branch as one `match` (throw-as-expression),
- `Flow\ETL\Formatter\ASCII\ASCIIValue::mb_str_pad()` — padding lengths hoisted above the `match`; `default => $result` reproduces the original no-default fall-through,
- `Flow\ETL\Adapter\Http\RequestEntriesFactory` — body-entry selection as a `match` assignment,
- `Flow\ETL\Adapter\Parquet\SchemaConverter::flowToParquet()` — 14-arm `match`; the `$nullable ? OPTIONAL : REQUIRED` ternary repeated in every case is hoisted into `$repetition`; `ListType`/`MapType` arms unwrap optionals through a small `unwrapOptional()` helper (repeated accessor calls defeat static-analysis narrowing inside a single arm expression),
- `Flow\ETL\Adapter\Seal\SchemaConverter::flowToSealField()` — same treatment,
- `Flow\Parquet\Writer` and `Flow\Parquet\Writer\ColumnChunkBuilderFactory` — compression/encoding validation as `match` with `condition ? throw : null` arms,
- `Flow\Parquet\Writer\PageBuilder\DictionaryBuilder` — previously a hybrid `switch` around nested `match`es, now a single nested `match`,
- `Flow\Parquet\Dremel\Validator\ColumnDataValidator` — nested validation `match`es; inner switches had no `default`, so the arms get an explicit `default => null` to preserve the silent pass-through, and `case null:` becomes a strict `null` arm.

Left as `switch` on purpose: `ThriftModel/*` and `CompactProtocol` (Thrift protocol field-id dispatch with multi-statement bodies / by-ref temporaries), `SnappyDecompressor` (hot loop with multi-variable assignments per case), `PageReader`, `FlowTelemetryBundle` (multi-statement DI configuration cases) and `PlainValuesPacker` (per-case `@var` assertions that static analysis relies on). None of these can become a `match` without restructuring beyond a mechanical conversion.

No public API changes. Every converted switch was checked for loose-vs-strict comparison hazards (subjects are enums, `::class` strings or `strtoupper()` results) and for missing-`default` fall-throughs (`match` throws `UnhandledMatchError` where `switch` falls through — explicit `default` arms preserve the original behavior).

## Verification & Benchmarks

Standard pipeline on 1M rows (`from_csv` → `batchSize(1000)` → `withEntry(concat)` → `filter(endsWith)` → write), PHP 8.3 + opcache, fresh process per run, medians of 3:

| Scenario | `1.x` | this PR | Δ |
|---|---|---|---|
| write CSV | 11.81 s | 11.67 s | within noise |
| write Parquet | 13.10 s | 13.30 s | within noise |

Peak memory unchanged (16/18 MB). **Output files are byte-identical** before and after the change (matching md5 for the 91 MB CSV and the 24.2 MB parquet output — the parquet variant deliberately routes through the converted `SchemaConverter`, `DictionaryBuilder` and per-value `ColumnDataValidator`). As expected for switch→match, this is a readability/strictness refactor, not a performance change — the benchmark is regression evidence.

Resolves: no related roadmap item — standalone modernization refactor

<h2>Change Log</h2>
<hr />
<div id="change-log">
  <h4>Added</h4>
  <ul id="added">
  </ul>
  <h4>Fixed</h4>
  <ul id="fixed">
  </ul>
  <h4>Changed</h4>
  <ul id="changed">
    <li>Replaced switch statements with match expressions across core ETL, parquet lib and http/parquet/seal adapters (no behavior change, outputs byte-identical)</li>
  </ul>
  <h4>Removed</h4>
  <ul id="removed">
  </ul>
  <h4>Deprecated</h4>
  <ul id="deprecated">
  </ul>
  <h4>Security</h4>
  <ul id="security">
  </ul>
</div>

### DIFF
--- a/src/adapter/etl-adapter-http/src/Flow/ETL/Adapter/Http/RequestEntriesFactory.php
+++ b/src/adapter/etl-adapter-http/src/Flow/ETL/Adapter/Http/RequestEntriesFactory.php
@@ -60,28 +60,20 @@ final class RequestEntriesFactory
             }
 
             if (!empty($requestBodyContent)) {
-                switch ($requestType) {
-                    case 'json':
-                        if (class_exists(JsonEntry::class)) {
-                            $decodedJson = type_array()->assert(json_decode(
+                $requestBodyEntry = match ($requestType) {
+                    'json' => class_exists(JsonEntry::class)
+                        ? new JsonEntry(
+                            'request_body',
+                            Json::fromArray(type_array()->assert(json_decode(
                                 $requestBodyContent,
                                 true,
                                 512,
                                 JSON_THROW_ON_ERROR,
-                            ));
-
-                            $requestBodyEntry = new JsonEntry('request_body', Json::fromArray($decodedJson));
-                        } else {
-                            $requestBodyEntry = string_entry('request_body', $requestBodyContent);
-                        }
-
-                        break;
-
-                    default:
-                        $requestBodyEntry = string_entry('request_body', $requestBodyContent);
-
-                        break;
-                }
+                            ))),
+                        )
+                        : string_entry('request_body', $requestBodyContent),
+                    default => string_entry('request_body', $requestBodyContent),
+                };
             }
         }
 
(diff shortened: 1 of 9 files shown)


norberttech:
I merged it mostly due to better DX, the perf improvement is hard to verify here

